PR #28, opened 2018-09-27
Add a very rough outline for a spec.

mkruisselbrink:
Doesn't actually specify any behavior yet, but is at least a start.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -16,23 +16,22 @@ Markup Shorthands: css no, markdown yes
 
 <pre class=link-defaults>
 spec:fetch; type:interface; text:ReadableStream
-spec:infra; type:dfn; for:/; text:set
 </pre>
 
 <pre class=anchors>
 </pre>
 
 <style>
-dl.domintro dt {
+.domintro dt {
     font-family: Menlo, Consolas, "DejaVu Sans Mono", Monaco, monospace;
 
     padding-top: 0.5em;
     padding-bottom: 1em;
 }
-dl.domintro dt a {
+.domintro dt a {
     color: inherit; border-bottom-style: none;
 }
-dl.domintro dt code {
+.domintro dt code {
     font-size: inherit;
 }
 </style>
@@ -57,7 +56,7 @@ Each [=/entry=] has an associated <dfn for=entry>name</dfn>.
 A <dfn lt="file|file entry">file entry</dfn> additionally consists of <dfn for="file entry">binary
 data</dfn> and a <dfn for="file entry">modification timestamp</dfn>.
 
-A <dfn lt="directory|directory entry">directory entry</dfn> additionally consists of a [=set=] of
+A <dfn lt="directory|directory entry">directory entry</dfn> additionally consists of a [=/set=] of
 <dfn for="directory entry">entries</dfn>. Each member is either a [=file=] or a [=directory=].
 
 The <dfn lt="parent|parent directory">parent directory</dfn> of a [=/entry=] |child| is the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -0,0 +1,505 @@
...
+Group: WICG
+Repository: wicg/writable-files
+Indent: 2
+Complain About: accidental-2119 yes, missing-example-ids yes
+Markup Shorthands: css no, markdown yes
+</pre>
+
+<pre class=link-defaults>
+spec:fetch; type:interface; text:ReadableStream
+spec:infra; type:dfn; for:/; text:set
```
inexorabletash: You should be able drop this if you use `[=/set=]` (at least, I was able to in one of my specs)

index.bs
```diff
@@ -0,0 +1,505 @@
...
+</div>
+
+### The {{FileSystemHandle/copyTo()}} method ### {#api-filesystemhandle-copyto}
+
+<div class=note><dl class=domintro>
+  : |newHandle| = await |handle| . {{FileSystemHandle/moveTo()|copyTo}}(await
+        |handle|.{{FileSystemHandle/getParent()}}, |newName|)
+  :: Creates a copy of the entry represented by |handle| with name |newName|.
+
+  : |newHandle| = await |handle| . {{FileSystemHandle/moveTo()|moveTo}}(|otherDir|)
```
inexorabletash: nit: s/moveTo/copyTo/ here and in the next dt

index.bs
```diff
@@ -0,0 +1,505 @@
...
+<pre class=link-defaults>
+spec:fetch; type:interface; text:ReadableStream
+spec:infra; type:dfn; for:/; text:set
+</pre>
+
+<pre class=anchors>
+</pre>
+
+<style>
+dl.domintro dt {
```
inexorabletash: I know you're replicating what other specs (mine?) do, but if we changed these selectors from `dl.domintro` to just `.domintro` then we could use `<div class="note domintro">` below and avoid the explicit `<dl>` tag.

index.bs
```diff
@@ -0,0 +1,505 @@
...
+
+# Accessing native filesystem # {#native-filesystem}
+
+## The {{Window/chooseFileSystemEntries()}} method ## {#api-choosefilesystementries}
+
+<xmp class=idl>
+enum ChooseFileSystemEntriesType { "openFile", "saveFile", "openDirectory" };
+
+dictionary ChooseFileSystemEntriesOptionsAccepts {
+  DOMString description;
```
inexorabletash: Make these `USVString`s, since they're probably going to get converted to UTF-8 at some point.

index.bs
```diff
@@ -0,0 +1,505 @@
...
+enum SystemDirectoryType {
+  "sandbox"
+};
+
+dictionary GetSystemDirectoryOptions {
+  required SystemDirectoryType type;
+};
+
+partial interface FileSystemDirectoryHandle {
+  static Promise<FileSystemDirectoryHandle> getSystemDirectory(GetSystemDirectoryOptions options);
```
inexorabletash: Hmmm... except for `sandbox` ISTM we'd want to hang this off Window instead, so operating systems can prompt the user (for permission, fallback directory, etc). Issue to track this?

index.bs
mkruisselbrink: Ah yes, I thought I tried that but must not have (and must have gotten confused by bikeshed showing multiple options for `/set` when I was just using `[=set=]`. But I guess the other one isn't exported so doesn't get picked up. Done.

index.bs
mkruisselbrink: Ah yes, that's nicer. Done.

index.bs

index.bs
mkruisselbrink: Good point. Done.

index.bs
mkruisselbrink: Yeah, I don't really have strong opinions on how this is exposed. Linked to the naming bikeshedding issue, as well as mentioned it perhaps being on Window. Although I don't see how the interface it is exposed on changes what operating systems can do wrt prompting?

inexorabletash: Just thinking if it's on Window then it's more obvious that you can't use it from serviceworkers where you can't prompt. But we need to solve that eventually and exceptions are a thing, so yeah - not a big reason to put it on Window.

PR #42, opened 2019-03-09
Fix link to the spec.

diegocr:
The link on README.md was still for the old repo name.


mkruisselbrink:
Thanks very much, unfortunately we renamed the repo again, so I just went ahead and made the change directly in master in dc64db87714357ad101b92c5631a14e4d4da7f35.

PR #43, opened 2019-03-09
Rename API in the spec document.

diegocr:


mkruisselbrink:
The spec is actually auto-generated from index.bs in the master branch. Directly editing index.html is not correct. I've updated index.bs in 4d25eefd75bb8eda5a9ad46c4977e7180ac57fcc, which will update gh-pages once travis is done with it.

Issue #10, opened 2016-10-09
Sandboxed File/Directory Path


feross:
One thing that could improve this API is an option to use a sandbox filesystem, similar to what the deprecated FileSystem API used to do.

Maybe we can add a `{sandbox: true}` or `{temp: true}` option that would store the file to a browser sandbox location without prompting the user.

Basically, everything would be the same, except the browser would not prompt the user for a file/directory location. Instead, the browser would just immediately return a sandboxed file/directory location.

Just set one option, and the browser will pick a path for you.

mkruisselbrink:
I do agree that this is something we might want to consider. It would essentially be something like the (chrome only) [`requestFileSystem`](https://www.w3.org/TR/2012/WD-file-system-api-20120417/#widl-LocalFileSystem-requestFileSystem-void-unsigned-short-type-unsigned-long-long-size-FileSystemCallback-successCallback-ErrorCallback-errorCallback) method, that returns a directory reference to a quota-managed and origin scoped sandboxed filesystem. Perhaps not in the initial version though, as it doesn't really help solve any new use cases. Although it might make testing easier (both for testing bits of this API as well as for people wanting to test code using these new writable file APIs).

jimmywarting:
I like the idea of a sandboxed Directory
But i would also want to control wether or not it should be temporary or persistent also. so i think both could be used. `{sandbox: true, temp: true}` or `{sandbox: true, persistent: true}`

using `temp` instead of `sandbox` is just confusing and is not the right word.

jespertheend:
The use case described here https://bugs.chromium.org/p/chromium/issues/detail?id=897313#c16 might require the use of a sandboxed file/direcotry.
I was wondering though, if we do end up adding some kind of way to open files with an external program from the browser, we might need to communicate to the user somehow that the opened file can still be accessed from the browser.

mkruisselbrink:
Closing this, since we do have a sandboxed filesystem in the spec now.

Issue #9, opened 2016-10-06
Super cookie issue


feross:
From the [Known Weirdness](https://github.com/WICG/writable-files/blob/master/EXPLAINER.md#known-weirdness) section:

> Sites could bill “save as…” like “download…” but then be able to retain access to it Could exacerbate super cookie issue from above 

Seems like this could be easily solved by clearing the file reference when cookies are cleared. Chrome already bundles a bunch of other settings into the clear cookies item: "Cookies and other site and plugin data"

<img width="290" alt="screen shot 2016-10-05 at 9 38 56 pm" src="https://cloud.githubusercontent.com/assets/121766/19140701/22d45c68-8b44-11e6-9644-3294e70d3ec2.png">

Sure this is another possible type of cookie, but it's not better at tracking users than existing cookies are.

drufball:
That's actually a really good point! Especially if we're requiring sites to persist the reference by storing it in IDB or something. We would get this behavior by default.

kenchris:
I guess you could also warn the user if a certain file/folder has been shared with more than N sites and ask for additional permission.

As Mike West long term would like to get rid of cookies, I wonder what he feels about this @mikewest

mikewest:
> I guess you could also warn the user if a certain file/folder has been shared with more than N sites and ask for additional permission.

It's not clear that this addresses the concern. I don't think folks are worried about sites sharing identifiers through the file, but sites using the file as an identifier with a lifetime distinct from the site's other identifiers (cookies, localStorage, etc).

Breaking the link between the site and the file when the rest of the site's data is cleared seems like a reasonable mitigation of that risk. I think I agree with @feross that this doesn't seem to provide any additional capabilities above and beyond cookies (aside, of course, from the file access in the first place).

> As Mike West long term would like to get rid of cookies, I wonder what he feels about this

We should get rid of cookies. But whatever we're left with will have the same characteristics insofar as we'd bundle it together with the rest of the site's data. Chrome, at least, tries to bundle these concepts up together, as it's unlikely to be helpful to ask users to make decisions about specific aspects of a site's data (localStorage vs IndexedDB, etc).

mkruisselbrink:
Yeah, we definitely want to make sure to revoke a sites access to all files when storage for the site is cleared. For file-picker selected files where you can only retain access by storing the reference in IDB we of course get that for free, but if we end up with more powerful APIs that let a site get access to certain native files without file pickers we might have to be more careful.

[...]

PR #50, opened 2019-05-14
Remove copy and move methods.

### LINKED ISSUE #52: Remove FileSystemHandle.copyTo()
`FileSystemHandle.copyTo()` can be implemented on top of `getFile()` / `getDirectory()` / `createWriter()` / `write()`, so it is not a new capability. A general-purpose implementation is non-trivial when considering recursively copying directories, attempting to restore the previous state on failures, and edge cases like symlinks and special files.

I think this method does not carry its weight in terms of capabilities added vs implementation effort required from browser developers, and I propose that we remove it for now.

mkruisselbrink:
These methods don't provide much value, since it is possible to get the
same behavior using the existing read and write APIs.

This fixes #52 and #53

### DIFF
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -192,8 +192,12 @@ const file_ref = await dir_ref.getFile('foo.js');
 // Get a subdirectory.
 const subdir = await dir_ref.getDirectory('bla', {createIfNotExists: true});
 
-// And you can possibly do stuff like move and/or copy files around.
-await file_ref.copyTo(dir_ref, 'new_name', {overwrite: true});
+// No special API to create copies, but still possible to do so by using
+// available read and write APIs.
+const new_file = await dir_ref.getFile('new_name', {create: true});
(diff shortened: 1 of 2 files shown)


jimmywarting:
So simply renaming a file requires you to read the hole file and write it to the new destination (essentially copying the hole file) and then deleting the original file 😞 

kind of wasteful operation and not so fast performant...

lgrahl:
Agree with @jimmywarting - totally sounds like something regrettable. Are you folks sure about this?

jimmywarting:
Say you want to rename a 4gb movie and only have 2 gb free space... problematic.

a-sully:
The PRs have unfortunately been in limbo for a while, but we hope to launch both [`move()`](https://github.com/whatwg/fs/pull/10) and [`remove()`](https://github.com/whatwg/fs/pull/9) relatively soon.

`move()` of file handles within the Origin Private File System _is_ currently supported (though not yet specified)

Issue #48, opened 2019-05-07
should be clear how API users can save without risk of corruption from partially completed writes


dbaron:
(Sorry, I think there's a bunch of terminology for this stuff that I think I used to know, but can't currently remember.  So I'm just describing the concepts without using the commonly-used names for them!)

When writing to files (e.g., for "save" functionality for documents, or for important configuration files), it's important not to overwrite a correct and complete copy of the file with a partial one, e.g., if the application crashes or the system loses power during the write.  So in many cases when writing to files, software will write to a separate temporary file, and then once that write has completed, replace the intended destination file with the temporary.  There are also more advanced ways of achieving the same goal.

The important piece is that if the use case is opening and later saving files, being able to do this with high reliability quickly becomes an important piece of it.

If this specification is going to advance, I think it should be clear what the path to achieving it in this model is.  It might be that the browser plays a significant role in doing this (likely preferable), or it might be that the API simply lets the application do it.  (However, if it does... the API needs to be structured in a way that doing it doesn't require any extra permission grants from the user, no matter what permission model the browser chooses to use, which seems like it might be a problem!)

So at the least I think the explainer should give examples of how to do this, and it seems like it might require changes in the feature set in the spec to handle this use case.

(Note that I got here from mozilla/standards-positions#154.)

mkruisselbrink:
Thanks for the feedback. I agree that this is an important use case that isn't sufficiently addressed. I create #51 with one possible solution for this.

pwnall:
@dbaron thank you very much for bringing up this important problem! I think that #51 solves it. The new default mode of operation is that files are written to a temporary location, and are atomically renamed into their final destination when the writer is closed.

Please open new issues if you see more opportunities for improvement!

Issue #37, opened 2018-11-29
Potentially Unwanted Programs Scan


shaqarava:
Currently, browsers perform a scan on files when downloaded. I'm not sure how this works under the hood but I know that some browsers have their own internal scanners. Also, that under most operating systems, the default virus/malware scanner is invoked in any newly downloaded path. Any such hooks should also be applied to paths newly created by this API.

mkruisselbrink:
This is definitely something we're considering. A few issues with it though: in chrome, for downloads, the way these checks work is that the file is first downloaded to a temporary file, then checked and finally moved/renamed to its actual destination. That flow won't really work if we want to support an efficient `pwrite` style API (#34) that lets a webapp modify individual bytes of files. Another problem with chrome's downloaded file checking is that it doesn't really work offline, and we explicitly do want this API to work offline as well.

So yeah, it would make sense to have some kind of safe browsing/virus scan checks as part of this, but I'm not sure I see a way of doing that while also keeping a reasonably performant API that actually solves the use cases we want to solve. Perhaps some balance is to only support `pwrite` style writing for "safe" file types (and definitely not for anything that might be executable), and restricting more dangerous file types to a "write this single Blob to the file" API.

DanielHerr:
Is there a reason Chrome or the operating system cannot scan the file after writing has been completed, or before the user attempts to open it?

mkruisselbrink:
"After writing has been completed" might not be very well defined with a pwrite style API. Doing a check after every individual write would probably cause unacceptable performance overhead, but doing the checks later after some delay/some amount of "idle" time could result in a bad user experience where chrome first tells the website that the write was successful (and the website probably then tells the user), but ultimately we end up somehow acting on detected possible malware.

Also unclear how we'd react on anything detected in such a way. Simply deleting the suspected malware files could be problematic, especially if a website is merely modifying an existing file on disk. Maybe we could try to undo the modifications, but that isn't really plausible either (without deep OS integration on certain types of filesystems only).

DanielHerr:
Are you aware of what happens with malware detected when the user selects an existing file to overwrite using the download functionality?

PR #51, opened 2019-05-15
Add options to createWriter() to deal with atomic writes.

mkruisselbrink:
This attempts to solve #37 by making it explicit that changes made to files
might not be reflected on disk until flush() or close() is called on the
writer.

This also introduces a "atomic" mode for FileSystemWriter where any changes
are written to a temporary file, and only on flush or close is that file
moved on top of the destination file. This should address #48.

Still to be decided if we actually want to implement the non-atomic mode at
first, as the semantics there aren't very clear. I.e. we might want to start with only
having an atomic mode.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -180,7 +180,7 @@ interface FileSystemFileHandle : FileSystemHandle {
 };
 </xmp>
 
-Issue: boolean's that default to "true" can be confusing (see comment in
+Issue: booleans that default to "true" can be confusing (see comment in
 https://heycam.github.io/webidl/#idl-dictionaries). We should try to come up with other names
 without ending up with double negations.
 
--- a/index.bs
+++ b/index.bs
@@ -206,7 +206,7 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method, when invoked, m
   : |writer| = await |fileHandle| . {{FileSystemFileHandle/createWriter()|createWriter}}({ {{FileSystemCreateWriterOptions/truncate}}: |truncate|, {{FileSystemCreateWriterOptions/atomic}}: true })
   :: Returns a {{FileSystemWriter}} that can be used to write to the file. Any changes made through
      |writer| won't be reflected in the file represented by |fileHandle| until its
-     {{FileSystemWriter/close()}} or {{FileSystemWriter/flush()}} methods are called.
+     {{FileSystemWriter/close()}} method is called.
      User agents try to ensure that no partial writes happen, i.e. the file represented by
      |fileHandle| will either contains its old contents or it will contain whatever data was written
      through |writer| up until one of these methods was called.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -261,11 +269,37 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method, when invoked, m
...
+     |writer| won't be reflected in the file represented by |fileHandle| until its
+     {{FileSystemWriter/close()}} or {{FileSystemWriter/flush()}} methods are called.
+     User agents try to ensure that no partial writes happen, i.e. the file represented by
+     |fileHandle| will either contains its old contents or it will contain whatever data was written
+     through |writer| up until one of these methods was called.
+
+     This is typically implemented by writing data to a temporary file, and only replacing the file
+     represented by |fileHandle| with the temporary file when the writer is closed.
+
+     If |truncate| is `true` or not specified, the temporary file starts out empty, otherwise the
```
oyiptong: Instead of a flag, how about an enum?

`fopen` provides the following options closest to what we want to achieve:
* `w` for truncate mode
* 'r+` for read/write, with the cursor at the beginning of the file

I like the idea of `update` to be the opposite of `truncate`.

If we want to contrast with `fopen` so there's no confusion, we can use slightly different enums:
```webidl

enum FileSystemWriterMode {"truncate", "update"};

dictionary FileSystemCreateWriterOptions {
  FileSystemWriterMode mode = "truncate";
  boolean atomic = true;
};
```

index.bs
```diff
@@ -461,6 +519,51 @@ steps:
 
 </div>
 
+### The {{FileSystemWriter/flush()}} method ### {#api-filesystemwriter-flush}
+
+<div class="note domintro">
+  : await |writer| . {{FileSystemWriter/flush()}}
```
pwnall: How about discussing flush() in a separate PR?

`flush()` for atomic files isn't intuitive to me. I'd expect that atomic=true implies that native apps on the system can only observe the file in the states marked by `flush()` and `close()` calls.  A browser would implement atomic files by writing to a temporary file, and atomically renaming to the destination file on `close()`. `flush()` would have to be implemented using the steps for `close()`, followed by copying the file into a temporary that receives future writes.

If the above is true, I'd rather that apps explicitly close the writer and create a new one.

`flush()` for in-place writers makes more sense, but feels like an optimization. Without flush, the app would have to close the writer and create a new writer, whereas `flush()` saves a bit of file descriptor churn.

In the in-place case, `flush()` adds a bunch of complexity.

1) Without `flush()`, we can say that the changes made to a file are guaranteed to be visible from native apps when the writer is closed (valid for all modes and options). This lets implementation grab an exclusive lock on the file, and release the lock after doing security checks. 
2) The name `flush()` is close enough to `FlushFileBuffers` that developers may assume it acts like `fsync()` and prevents data loss on power failure. While we'll need to discuss guarantees in this spec anyway, I'd prefer that we stay away from all that `flush()` implies until we have a use case where creating a new writer doesn't cut it.

mkruisselbrink: Good point, I'll remove flush for now, and I'm mostly convinced we probably don't need it anyway.

index.bs
mkruisselbrink: I went with victor's suggested names for now.

PR #55, opened 2019-05-16
Change the remove operation to be part of a directory handle.

mkruisselbrink:
Before each handle had a remove method that would remove the entry
represented by the handle. This changes it to instead have a removeEntry
method on a directory handle that will remove a child of that directory.

This means it is no longer possible to delete an individual file or
directory without having a handle to its parent, but otherwise feels
like a slightly nicer API shape to me.

This fixes#54

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -74,7 +74,6 @@ interface FileSystemHandle {
   readonly attribute boolean isDirectory;
   readonly attribute USVString name;
 
-  Promise<void> remove();
 
   Promise<PermissionState> queryPermission(optional FileSystemHandlePermissionDescriptor descriptor);
   Promise<PermissionState> requestPermission(optional FileSystemHandlePermissionDescriptor descriptor);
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -291,7 +280,7 @@ interface FileSystemDirectoryHandle : FileSystemHandle {
 
   Promise<sequence<USVString>?> resolve(FileSystemHandle handle);
 
-  Promise<void> removeRecursively();
+  Promise<void> removeEntry(USVString name, optional FileSystemRemoveOptions options);
```
pwnall: While I personally like `remove`, the web platform seems to have settled on `delete` for this operation.

So, WDYT of `delete`, or `deleteEntry`?

Also fine to leave the name as-is and let the TAG decide when we submit a review.

mkruisselbrink: I think I'll leave as is for now. Which web platform APIs were you referring to btw? I guess CacheStorage indeed uses `delete`, but for example DOM has `removeChild` and `removeAttribute`

pwnall:
I was thinking primarily of https://github.com/WICG/cookie-store/issues/46. I agree we shouldn't block the API improvements in this PR on the name, though. I'll open an issue to track the name and let folks chime in there.

guest271314:
At Chromium 77 `removeRecursively()` is a property of `FileSystemDirectoryHandle`, not `removeEntry()`. The specification (Draft Community Group Report, 21 June 2019)  does not contain any language referencing `removeRecursively()`. A note describing the intention to change the name of the including a link pointing to this PR in the specification would be helpful. Can file a separate PR for that putpose if that is necessary.

Issue #19, opened 2018-08-02
Why not use ReadableStream/WritableStream/WritableStream*Writer


kenchris:
Instead of the FileReader and FileWriter.

It could potentially be specialized versions, or at least the new objects could share API

If there are obvious reasons why these cannot be used, I think the explainer should say so.

Of course FileReader already exists, but FileWriter doesn't to my knowledge, so it would be good to figure out what is the best API going forward.

Currently you have `createWriter` and `WritableStream` has `getWriter` returning a `WritableStreamDefaultWriter`:

https://streams.spec.whatwg.org/#default-writer-class

The FileReader also doesn't support streaming of content, unless you read manually by using `.slice()`

mkruisselbrink:
For reading, see #2 and w3c/FileAPI#40; i.e. I'm not against adding something to File/Blob to make it easier to use a stream to read from a file, but it's already pretty easy to create a ReadableStream out of a Blob (and no, that doesn't require using slice).

For writing, this can't just use WritableStream at least, as that wouldn't support seeking/random access. But perhaps it would indeed make sense to optionally let you treat a file reference as a WritableStream. Not quite sure what that would look like. Also the explainer currently (somewhat on purpose) completely glosses over things like locking files, if multiple writers should be able to write to the same file at the same time, etc.

kenchris:
Yes, a WritableStream would lock by default. At least the FileWriter could attempt to use the same method names for the same things.

domenic:
I'd like to bump this a bit. The current shape of the spec is pretty nice: you get a `FileSystemWriter` which has a `write(position, data)`, `asWritableStream()`, `truncate()`, and `close()` method. But I think it could be even nicer.

What if instead you had a `WritableFileStream` subclass of `WritableStream`, which added the seek and truncate operations? Then we'd stay with a single paradigm for writing, instead of introducing a new one with a converter method to get to the standard one. You'd probably want to also create a `WritableFileStreamDefaultWriter` subclass of `WritableStreamDefaultWriter`.

Then the code would look something like

```js
const stream = await fileHandle.createWritable();

// All I want to do is truncate:
await stream.truncate(1000);

// Or I want to do more complicated stuff:
const writer = stream.getWriter();
writer.seek(10);
writer.write(uint8Array);
writer.truncate(100);

// Or I want to compose with readablestream:
await readable.pipeTo(stream);
```

What do you think?

mkruisselbrink:
[...]

PR #62, opened 2019-06-11
Replace createWriter, FileSystemWriter with createWritable, WritableFileStream

oyiptong:
This adds specs for `WritableFileStream` and the factory and options.
It links to `WritableStreams` related specs, with an inheritance
relationship.

***
<a href="https://pr-preview.s3.amazonaws.com/oyiptong/native-file-system/pull/62.html" title="Last updated on Mar 10, 2020, 4:45 PM UTC (890e901)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/62/2cc2ae0...oyiptong:890e901.html" title="Last updated on Mar 10, 2020, 4:45 PM UTC (890e901)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -190,14 +190,14 @@ The <dfn method for=FileSystemHandle>requestPermission(|descriptor|)</dfn> metho
 ## The {{FileSystemFileHandle}} interface ## {#api-filesystemfilehandle}
 
 <xmp class=idl>
-dictionary FileSystemCreateWriterOptions {
+dictionary FileSystemCreateWritableOptions {
   boolean keepExistingData = false;
 };
 
 [Exposed=(Window,Worker), SecureContext, Serializable]
 interface FileSystemFileHandle : FileSystemHandle {
   Promise<File> getFile();
-  Promise<FileSystemWriter> createWriter(optional FileSystemCreateWriterOptions options = {});
+  Promise<FileSystemWritableFileStream> createWritable(optional FileSystemCreateWritableOptions options);
 };
 </xmp>
 
@@ -222,27 +222,29 @@ The <dfn method for=FileSystemFileHandle>getFile()</dfn> method, when invoked, m
 
 </div>
 
-### The {{FileSystemFileHandle/createWriter()}} method ### {#api-filesystemfilehandle-createwriter}
+### The {{FileSystemFileHandle/createWritable()}} method ### {#api-filesystemfilehandle-createwritable}
+
+Advisement: In the Origin Trial as available in Chrome 82, createWritable replaces the createWriter method.
 
 <div class="note domintro">
-  : |writer| = await |fileHandle| . {{FileSystemFileHandle/createWriter()}}
-  : |writer| = await |fileHandle| . {{FileSystemFileHandle/createWriter()|createWriter}}({ {{FileSystemCreateWriterOptions/keepExistingData}}: true/false })
-  :: Returns a {{FileSystemWriter}} that can be used to write to the file. Any changes made through
-     |writer| won't be reflected in the file represented by |fileHandle| until its
-     {{FileSystemWriter/close()}} method is called.
+  : |stream| = await |fileHandle| . {{FileSystemFileHandle/createWritable()}}
+  : |stream| = await |fileHandle| . {{FileSystemFileHandle/createWritable()|createWritable}}({ {{FileSystemCreateWritableOptions/keepExistingData}}: true/false })
+  :: Returns a {{FileSystemWritableFileStream}} that can be used to write to the file. Any changes made through
+     |stream| won't be reflected in the file represented by |fileHandle| until its
+     {{FileSystemWritableFileStream/close()}} method is called.
      User agents try to ensure that no partial writes happen, i.e. the file represented by
      |fileHandle| will either contains its old contents or it will contain whatever data was written
-     through |writer| up until {{FileSystemWriter/close()}} was called.
+     through |stream| up until {{FileSystemWritableFileStream/close()}} was called.
 
      This is typically implemented by writing data to a temporary file, and only replacing the file
-     represented by |fileHandle| with the temporary file when the writer is closed.
+     represented by |fileHandle| with the temporary file when the writable filestream is closed.
 
-     If {{FileSystemCreateWriterOptions/keepExistingData}} is `false` or not specified,
+     If {{FileSystemCreateWritableOptions/keepExistingData}} is `false` or not specified,
      the temporary file starts out empty,
      otherwise the existing file is first copied to this temporary file.
 </div>
 
-Issue: There has been some discussion around and desire for a "inPlace" mode for createWriter (where
+Issue: There has been some discussion around and desire for a "inPlace" mode for createWritable (where
 changes will be written to the actual underlying file as they are written to the writer, for example
 to support in-place modification of large files or things like databases). This is not currently
 implemented in Chrome. Implementing this is currently blocked on figuring out how to combine the
@@ -250,7 +252,7 @@ desire to run malware checks with the desire to let websites make fast in-place 
 existing large files.
 
 <div algorithm>
-The <dfn method for=FileSystemFileHandle>createWriter(|options|)</dfn> method, when invoked, must run these steps:
+The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method, when invoked, must run these steps:
 
 1. TODO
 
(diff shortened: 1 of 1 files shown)


oyiptong:
@domenic, what do you think of this PR as a response to #19?

In particular, what do you think of the initialization of `position` in `createWritable()`. This would be used instead of a `seek()` method on the `WritableFileStream`.

index.bs
```diff
@@ -403,107 +410,100 @@ these steps:
 
 </div>
 
-## The {{FileSystemWriter}} interface ## {#api-filesystemwriter}
+## The {{FileSystemWritableFileStream}} interface ## {#api-filesystemwritablefilestream}
 
 <xmp class=idl>
 [Exposed=(Window,Worker), SecureContext]
-interface FileSystemWriter {
+interface FileSystemWritableFileStream {
```
mkruisselbrink: this should actually inherit from WritableStream, right?

index.bs
```diff
@@ -168,15 +168,16 @@ The <dfn method for=FileSystemHandle>requestPermission(|descriptor|)</dfn> metho
 ## The {{FileSystemFileHandle}} interface ## {#api-filesystemfilehandle}
 
 <xmp class=idl>
-dictionary FileSystemCreateWriterOptions {
+dictionary FileSystemCreateWritableOptions {
+  unsigned long long position = 0;
```
mkruisselbrink: I think I'd prefer having this be something that is passed to getWriter, rather than have it be a part of creating the stream. I.e. have FileSystemWritableStream.getWriter take an (optional?) position that starts the writer off at that position. This really is a very different option than the other options that can be passed when creating a writer thing, and it also seems more consistent to me if the position is something that you pass in to all the methods in FileSystemWritableStream, rather than have some use some kind of implied position and some have an explicit position.

Having said that, I guess we'd still need some kind of implicit default position for when no offset is specified for getWriter, or when you pipe another stream into this stream... In that case I think having a seek() on the stream itself would still be better than passing in a position here.

But curious what domenic thinks.

index.bs
```diff
@@ -403,107 +410,100 @@ these steps:
...
 
-Issue(19): The functionality described by this method will likely stay, but the exact shape of
-the method is still in flux even more than the rest of this specification.
+### The {{FileSystemWritableFileStream/write()}} method ### {#api-filesystemwritablefilestream-write}
 
 <div class="note domintro">
-  : let |stream| = |writer| . {{FileSystemWriter/asWritableStream()}}
-  :: Returns a {{WritableStream}} that can be used to write data into the file, starting
-     at the beginning of the file.
+  : await |writableFStream| . {{FileSystemWritableFileStream/write()|write}}(|position|, |data|)
```
mkruisselbrink: do we want this (and I suppose possibly truncate below as well) to update the position at which future getWriter() calls will start writing?

index.bs
```diff
@@ -403,107 +410,100 @@ these steps:
...
 
-### The {{FileSystemWriter/close()}} method ### {#api-filesystemwriter-close}
+### The {{FileSystemWritableFileStream/close()}} method ### {#api-filesystemwritablefilestream-close}
 
 <div class="note domintro">
-  : await |writer| . {{FileSystemWriter/close()}}
-  :: First flushes any data written so far to disk, and then closes the writer.
-     If |writer|'s [=atomic flag=] is set, no changes will
+  : await |writableFStream| . {{FileSystemWritableFileStream/close()}}
+  :: This method is similar to {{WritableStreamDefaultWriter/close()|WritableStreamDefaultWriter's close()}}
```
mkruisselbrink: should just calling close() on the Writer have the same behavior? I imagine it should, as otherwise it would be kind of weird that the only way to actually get your changes made using a writer to persist on disk would be to first call `releaseLock` on the writer, and then call close on the stream. (I at least imagine we would want write/truncate/close on the stream to only work when the stream isn't locked, and to itself keep the stream locked during the duration of the operation).

index.bs
```diff
@@ -168,15 +168,16 @@ The <dfn method for=FileSystemHandle>requestPermission(|descriptor|)</dfn> metho
 ## The {{FileSystemFileHandle}} interface ## {#api-filesystemfilehandle}
 
 <xmp class=idl>
-dictionary FileSystemCreateWriterOptions {
+dictionary FileSystemCreateWritableOptions {
+  unsigned long long position = 0;
   boolean inPlace = false;
   boolean keepExistingData = false;
```
mkruisselbrink: I wonder if "keepExistingData" should be more of an enum like you suggested so we can have append/truncate/overwrite modes, where append would start at the end of the file, and truncate/overwrite would start at the end of the file... That combined with possibly a seek() method (and no other explicit initial position), or the ability to pass in a position to getWriter almost feels like it might be the best of both worlds. I suppose the only thing missing if we don't have an explicit initial position or seek method is the ability to pipe into the middle of an existing file, but that perhaps is an okay compromise?

index.bs
oyiptong: correct. Fixed.

index.bs
oyiptong: I personally prefer `seek()` on the stream as it's more explicit. However, we still have the issue of the implicit position on the stream.

index.bs
oyiptong: Good question. Given that this takes an explicit `position`, my take on it would be that it wouldn't, but I look forward to the discussion on it.

After calling `write()` with an explicit position, let's consider two possible behaviors:
1. the implicit `position` is changed to the last byte just written
2. the implicit `position` is unchanged

To me, number 2 seems to be the least surprising, because the use of an explicit position option seems to imply eschewing the use of stream-like functionality, perhaps temporarily, for explicit control on the start position.

index.bs
oyiptong: I think it should.

index.bs
domenic: I don't have strong feelings here. I think one could argue that this is a peer to inPlace, but I'm not sure (and inPlace doesn't seem to have a domintro for me to skim to try to understand it).

I do think you need to store a position on the stream if you make it part of getWriter(); i.e. you should finish with one writer, then getting a new one should pick up where you left off. And at that point it's not much different then adding seek(). You could even say that seek() just adjusts where the next getWriter() will start from.

If you instead make this an argument to the stream, then I could see getWriter() always starting at the specified position... Basically the stream represents "the file at position X", and you can keep overwriting that segment multiple times by getting multiple writers.

index.bs
```diff
@@ -403,107 +410,100 @@ these steps:
...
 </div>
 
-## The {{FileSystemWriter}} interface ## {#api-filesystemwriter}
+## The {{FileSystemWritableFileStream}} interface ## {#api-filesystemwritablefilestream}
 
 <xmp class=idl>
 [Exposed=(Window,Worker), SecureContext]
-interface FileSystemWriter {
+interface FileSystemWritableFileStream : WritableStream {
   Promise<void> write(unsigned long long position, (BufferSource or Blob or USVString) data);
```
domenic: You shouldn't define write() and close(); users should call those via the writer.

To customize the behavior of the writer, instead you need to define (in the definition of createWritable) what the underlying sink algorithms are. See https://fetch.spec.whatwg.org/#concept-construct-readablestream and its call sites for some rough guidance (although that's for readable; for writable you'd be using https://streams.spec.whatwg.org/#create-writable-stream).

index.bs
```diff
@@ -403,107 +410,100 @@ these steps:
 
 </div>
 
-## The {{FileSystemWriter}} interface ## {#api-filesystemwriter}
+## The {{FileSystemWritableFileStream}} interface ## {#api-filesystemwritablefilestream}
 
 <xmp class=idl>
 [Exposed=(Window,Worker), SecureContext]
-interface FileSystemWriter {
+interface FileSystemWritableFileStream : WritableStream {
```
domenic: If you're changing the return value of getWriter(), you need to overwrite it.

index.bs
```diff
@@ -403,107 +410,100 @@ these steps:
...
-<div algorithm>
-The <dfn method for=FileSystemWriter>write(|position|, |data|)</dfn> method, when invoked, must run
-these steps:
+The {{FileSystemWritableFileStream}} can be initialized with a |position|. When obtaining a writer through
+{{WritableStream/getWriter()|WritableStream's getWriter()}}, and using the returned
+{{WritableStreamDefaultWriter/write()|WritableStreamDefaultWriter's write()}} method, this position will be
+advanced based on the number of bytes written through this stream object.
 
-1. TODO
+{{WritableStream/getWriter()|WritableStream's getWriter()}} method returns a raw {{WritableStreamDefaultWriter}}.
```
domenic: What does "raw" mean? How is the writer initialized?

index.bs
domenic: See my above comments for how to achieve that, by defining underlying sink algorithms.

index.bs
oyiptong: updated to use a `seek` method.

index.bs
oyiptong: updated `write` to not take `position` as a positional argument.

index.bs
oyiptong: Behavior now defined in WritableStream.

index.bs
oyiptong: We do now have a `seek` method without an explicit initial position. The enum for creating the stream is a good idea.

Tracking in #141

index.bs
oyiptong: What do you think of defining `write()` and `truncate()`, so that `FileSystemWritableFileStream` can be used directly?

For instance, one could do:

```js
const stream = await handle.createWritable();
await stream.write(new Blob(['foo']); // New method, peer to `seek` and `truncate`.
await stream.close(); // Overrided method from WritableStream.
```

At the same time, the underlying sink algorithms will be implemented to accept `WriteParams` to execute commands, or to take byte-like data.

index.bs
oyiptong: getWriter() will now return an instance of `WritableStreamDefaultWriter`

index.bs
oyiptong: changed the wording.

oyiptong:
Closes #19

PR #160, opened 2020-03-12
Misc updates

mkruisselbrink:
Primarily adding resolve/isSameEntry back to the spec to match our implementation, but also some minor other changes.

***

### :boom: Error: write EPROTO 140222018242432:error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version:../deps/openssl/openssl/ssl/s23_clnt.c:772:
 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 16, 2020, 6:51 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2FWICG%2Fnative-file-system%2Fpull%2F160%2F48e383f.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2FWICG%2Fnative-file-system%2Fpull%2F160.html)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/native-file-system%23160.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -16,6 +16,7 @@ Markup Shorthands: css no, markdown yes
 
 <pre class=link-defaults>
 spec:fetch; type:interface; text:ReadableStream
+spec:webidl; type:dfn; text:resolve
 </pre>
 
 <pre class=anchors>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -50,19 +51,51 @@ This provides similar functionality as earlier drafts of the
...
-<dfn for="directory entry">entries</dfn>. Each member is either a [=file=] or a [=directory=].
+<dfn for="directory entry">children</dfn>, which are themselves [=/entries=]. Each member is either a [=file=] or a [=directory=].
 
-Issue: TODO: Explain how entries map to files on disk (multiple entries can map to the same file or
+[=/Entries=] can (but don't have to) be backed by files on the systems native file system,
+as such it is possible for the [=binary data=], [=modification timestamp=],
+and [=children=] of entries to be modified by applications outside of this specification.
+Exactly how external changes are reflected in the data structures defined by this specification,
+as well as how changes made to the data structures defined here are reflected externally
+is left up to individual user-agent implementations.
```
pwnall: I see two possible meanings here.

1) The logical structures in this spec are mapped to filesystems in implementation-defined ways. I suppose this is reasonable. We could add an example of concepts map to POSIX filesystems. This covers every operating system I'm aware of -- Windows implements a subset of POSIX.

2) The exact sequence in which filesystem modifications reach the file system are implementation-dependent. This seems reasonable, with the caveat that we do have some guarantees for atomic writes -- no visible filesystem changes until time T1, all changes (file data, size) visible by time T2.

index.bs
```diff
@@ -50,19 +51,51 @@ This provides similar functionality as earlier drafts of the
...
+
+An [=/entry=] |a| is <dfn for="entry">the same as</dfn> an [=/entry=] |b| if |a| is equal to |b|, or
+if |a| and |b| are backed by the same file or directory on the native file system.
+
+Issue: TODO: Explain better how entries map to files on disk (multiple entries can map to the same file or
 directory on disk but doesn't have to map to any file on disk).
```
pwnall: (not introduced in this commit)

"but an entry doesn't have to map to..."?

The previous clause's subject is "multiple entries" (plural). This sentence probably needs a different subject, or a singular/plural agreement fix (doesn't -> don't).

index.bs
mkruisselbrink: I think I kind of mean both meanings, but it's definitely a bit clumsy as written. Not sure how to improve, although I suspect I'll be tweaking/modifying this text anyway as I try to actually better define individual algorithms. I expect we might end up with non-normative notes at the various algorithms explaining some of the things (like atomic writes)...

PR #161, opened 2020-03-17
More formally specify the streams integration.

mkruisselbrink:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/161.html" title="Last updated on Mar 18, 2020, 7:12 PM UTC (1255d79)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/161/449ec17...1255d79.html" title="Last updated on Mar 18, 2020, 7:12 PM UTC (1255d79)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -59,7 +59,7 @@ A <dfn lt="file|file entry">file entry</dfn> additionally consists of <dfn for="
 data</dfn> and a <dfn for="file entry">modification timestamp</dfn>.
 
 A <dfn lt="directory|directory entry">directory entry</dfn> additionally consists of a [=/set=] of
-<dfn for="directory entry">children</dfn>, which are themselves [=/entries=]. Each member is either a [=file=] or a [=directory=].
+<dfn for="directory entry">children</dfn>, which are themselves [=/entries=]. Each member is either a [=/file=] or a [=directory=].
 
 [=/Entries=] can (but don't have to) be backed by files on the systems native file system,
 as such it is possible for the [=binary data=], [=modification timestamp=],
@@ -290,11 +290,10 @@ Advisement: In the Origin Trial as available in Chrome 82, createWritable replac
   : |stream| = await |fileHandle| . {{FileSystemFileHandle/createWritable()}}
   : |stream| = await |fileHandle| . {{FileSystemFileHandle/createWritable()|createWritable}}({ {{FileSystemCreateWritableOptions/keepExistingData}}: true/false })
   :: Returns a {{FileSystemWritableFileStream}} that can be used to write to the file. Any changes made through
-     |stream| won't be reflected in the file represented by |fileHandle| until its
-     {{FileSystemWritableFileStream/close()}} method is called.
+     |stream| won't be reflected in the file represented by |fileHandle| until the stream has been closed.
      User agents try to ensure that no partial writes happen, i.e. the file represented by
      |fileHandle| will either contains its old contents or it will contain whatever data was written
-     through |stream| up until {{FileSystemWritableFileStream/close()}} was called.
+     through |stream| up until the stream has been closed.
 
      This is typically implemented by writing data to a temporary file, and only replacing the file
      represented by |fileHandle| with the temporary file when the writable filestream is closed.
@@ -314,7 +313,12 @@ modifications to existing large files.
 <div algorithm>
 The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method, when invoked, must run these steps:
 
-1. TODO
+1. Issue: TODO: Check permissions.
+1. Let |stream| be the result of [=create a new FileSystemWritableFileStream|creating a new FileSystemWritableFileStream=]
+   for <b>[=this=]</b>'s [=FileSystemHandle/entry=].
+1. If |options|.{{FileSystemCreateWritableOptions/keepExistingData}} is `true`:
+   1. Set |stream|.[=[[buffer]]=] to a copy of <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=file entry/binary data=].
+1. Return [=a promise resolved with=] |stream|.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
@domenic, @ricea would one of you be able to take a look at this from the streams side of things? This tries to specify what we have currently implemented.

index.bs
```diff
@@ -535,101 +539,233 @@ dictionary WriteParams {
...
-     If {{WriteParams/type|type}} is {{write}}, {{WriteParams/position|position}} is an <span class=allow-2119>optional</span> attribute, {{WriteParams/data|data}} is a <span class=allow-2119>required</span> attribute.
+<div algorithm>
+To <dfn>create a new FileSystemWritableFileStream</dfn> given a [=file entry=] `file`, perform the following steps:
+
+1. Let |stream| be a new {{FileSystemWritableFileStream}}.
+1. Perform [$InitializeWritableStream$](|stream|)
+1. Set |stream|.[=FileSystemWritableFileStream/[[file]]=] to [=file entry=].
+1. Let |controller| be a new {{WritableStreamDefaultController}}.
+1. Let |startAlgorithm| be an algorithm that returns `undefined`.
+1. Let |writeAlgorithm| be an algorithm which takes a |chunk| argument
```
ricea: |writeAlgorithm| has to return a promise. It's fine to add "and returns a promise resolved with undefined." here.

index.bs
```diff
@@ -535,101 +539,233 @@ dictionary WriteParams {
...
+   and runs the [=write a chunk=] algorithm with |stream| and |chunk|.
+1. Let |closeAlgorithm| be the following steps:
+   1. Issue: TODO: Check permissions.
+   1. Set |stream|.[=[[file]]=]'s [=file entry/binary data=] to |stream|.[=[[buffer]]=].
+      If that throws an exception, return [=a promise rejected with=] that exception.
```
ricea: Also needs to return a promise on success.

index.bs
```diff
@@ -535,101 +539,233 @@ dictionary WriteParams {
...
+
+      Note: it is expected that this atomically updates the contents of the file on disk
+      being written to.
+1. Let |abortAlgorithm| be an algorithm that returns `undefined`.
```
ricea: The return value has to be strictly a promise, so it needs to be "an algorithm that returns a promise resolved with undefined".

index.bs
```diff
@@ -535,101 +539,233 @@ dictionary WriteParams {
...
+  :: Updates the current file cursor offset the |position| bytes from the top of the file.
+</div>
+
+<div algorithm>
+The <dfn method for=FileSystemWritableFileStream>seek(|position|)</dfn> method, when invoked, must run these
 steps:
 
-1. TODO
+1. Let |writer| be the result of calling [$AcquireWritableStreamDefaultWriter$](<b>[=this=]</b>).
+1. Let |result| be [$WritableStreamDefaultWriterWrite$](|writer|, <code>{type: {{WriteCommandType/"seek"}}, position: |position|}</code>).
```
ricea: You should keep in mind that when this object is converted to a `WriteParams` dictionary later, it will access the global`Object.prototype.size` and `Object.prototype.data`. There may be getters on these properties that execute arbitrary JavaScript. 

It's probably worth including a non-normative note about this, as it may have security implications for the implementor.

Alternatively, you could explicitly set `size` and `data` to undefined here, to avoid the possible side-effects.

index.bs
```diff
@@ -535,101 +539,233 @@ dictionary WriteParams {
   (BufferSource or Blob or USVString)? data;
 };
 
+typedef (BufferSource or Blob or USVString or WriteParams) FileSystemWriteChunkType;
+
 [Exposed=(Window,Worker), SecureContext]
 interface FileSystemWritableFileStream : WritableStream {
-  Promise<void> write((BufferSource or Blob or USVString or WriteParams) data);
+  Promise<void> write(FileSystemWriteChunkType data);
```
domenic: Hmm. If specs are going to be subclassing us to add convenience methods like this anyway, I wonder if we should add this to WritableStream.prototype...

Alternately this could just be a convenient wrapper for the `{ type: "write" }` variant, and not accept all the different chunk types.

index.bs
```diff
@@ -535,101 +539,233 @@ dictionary WriteParams {
...
+A {{FileSystemWritableFileStream}} has an associated <dfn for=FileSystemWritableFileStream>\[[file]]</dfn> (a [=file entry=]).
 
-The underlying sink's write method, and therefore {{WritableStreamDefaultWriter/write()|WritableStreamDefaultWriter's write()}} method, will accept byte-like data or {{WriteParams}} as input.
+A {{FileSystemWritableFileStream}} has an associated <dfn for=FileSystemWritableFileStream>\[[buffer]]</dfn> (a [=byte sequence=]).
+It is initially empty.
 
-The {{FileSystemWritableFileStream}} has a file position cursor initialized at byte offset 0 from the top of the file. When using {{FileSystemWritableFileStream/write()|write()}} or by using WritableStream capabilities through the {{WritableStreamDefaultWriter/write()|WritableStreamDefaultWriter's write()}} method, this position will be advanced based on the number of bytes written through the stream object.
+Note: This buffer can get arbitrarily large, so it is expected that implementations will not keep this in memory,
+but instead use a temporary file for this.
```
domenic: Ah, this helps explain why createWritable() returns a promise. It might be good to note that all access to [[buffer]] is done in promise-returning methods, so even though it looks sync in the spec, implementations can make it async.

mkruisselbrink: The other reason createWritable() returns a promise is the "TODO: Check permissions.", in particular if the user doesn't already have write permission trying to create a writable will prompt for that permission.

Anyway, I agree that adding such a sentence about [[buffer]] is helpful, so I'll do that.

index.bs
mkruisselbrink: Yeah, not sure what the best long-term shape for this API is; this is what we've currently implemented, but I'd be totally up for changes before we ship this outside of an origin trial if we think a different API makes more sense.

index.bs
ricea: It's pretty arbitrary. Mostly it was just easier to spec that way. One justification is that other specifications rarely need to do anything in the start method, so it's easier for them to just return undefined.

PR #164, opened 2020-03-26
Start fleshing out permissions integration.

mkruisselbrink:
In the future we should figure out how to integrate this better with the
permissions specification, but for now this documents at least what is
implemented in chrome.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/164.html" title="Last updated on Mar 30, 2020, 11:43 PM UTC (bec294c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/164/532084d...bec294c.html" title="Last updated on Mar 30, 2020, 11:43 PM UTC (bec294c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -12,7 +12,6 @@ Repository: wicg/native-file-system
 Indent: 2
 Complain About: accidental-2119 yes, missing-example-ids yes
 Markup Shorthands: css no, markdown yes
-Include Mdn Panels: no
 </pre>
 
 <pre class=link-defaults>
@@ -73,7 +72,7 @@ if it was returned by {{chooseFileSystemEntries()}} or {{getSystemDirectory()}},
 will have a parent in all other cases.
 
 [=/Entries=] can (but don't have to) be backed by files on the systems native file system,
-as such it is possible for the [=binary data=], [=modification timestamp=],
+so it is possible for the [=binary data=], [=modification timestamp=],
 and [=children=] of entries to be modified by applications outside of this specification.
 Exactly how external changes are reflected in the data structures defined by this specification,
 as well as how changes made to the data structures defined here are reflected externally
@@ -109,11 +108,6 @@ run the following steps:
 
 ## Permissions ## {#permissions}
 
-Note: By default entries inherit their permissions from their parent. If an entry doesn't
-have a parent, permissions are by default granted, and requesting permissions is a no-op.
-Entries returned by {{chooseFileSystemEntries()}} override these steps, providing stricter
-access control for those entries.
-
 Issue: Figure out some way to integrate this better with the permissions API [[permissions]],
 for example by making the entry be a part of the PermissionDescriptor.
 
@@ -122,7 +116,13 @@ An [=/entry=] |entry| has associated <dfn for=entry>query permission steps</dfn>
 which take a {{FileSystemHandlePermissionDescriptor}} and return a {{PermissionState}}.
 Unless specified otherwise, these steps are:
 1. Let |parent| be |entry|'s [=entry/parent=].
-1. If |parent| is null, return {{PermissionState/"granted"}}.
+1. Assert: |parent| is not null.
+
+   Note: [[#native-file-system-permissions]] overrides these steps for entries returned by
+   {{chooseFileSystemEntries()}}. Additionally [[#special-filesystem-concepts]] overrides
+   these steps for entries returned by {{FileSystemDirectoryHandle/getSystemDirectory()}}.
+   All other entries always have a parent.
+
 1. Return the result of running |parent|'s [=query permission steps=].
 
 Note: These steps return the current state synchronously, however these steps are only
@@ -381,18 +381,19 @@ The <dfn method for=FileSystemFileHandle>createWritable(|options|)</dfn> method,
 
 1. Let |result| be [=a new promise=].
 1. Run the following steps [=in parallel=]:
-  1. Run <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=request permission steps=] given
+  1. Let |entry| be <b>[=this=]</b>'s [=FileSystemHandle/entry=].
+  1. Run |entry|'s [=request permission steps=] given
      «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `true`&nbsp;]».
      If that throws an exception, [=reject=] |result| with that exception and abort.
   1. Let |permissionStatus| be the result of running
-     <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=query permission steps=] given
+     |entry|'s [=query permission steps=] given
      «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `true`&nbsp;]».
   1. If |permissionStatus| is not {{PermissionState/"granted"}},
      reject |result| with a {{NotAllowedError}} and abort.
   1. Let |stream| be the result of [=create a new FileSystemWritableFileStream|creating a new FileSystemWritableFileStream=]
-     for <b>[=this=]</b>'s [=FileSystemHandle/entry=].
+     for |entry|.
   1. If |options|.{{FileSystemCreateWritableOptions/keepExistingData}} is `true`:
-    1. Set |stream|.[=[[buffer]]=] to a copy of <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=file entry/binary data=].
+    1. Set |stream|.[=[[buffer]]=] to a copy of |entry|'s [=file entry/binary data=].
   1. [=/Resolve=] |result| with |stream|.
 1. Return |result|.
 
@@ -472,16 +473,17 @@ must run these steps:
 
 1. Let |result| be [=a new promise=].
 1. Run the following steps [=in parallel=]:
+  1. Let |entry| be <b>[=this=]</b>'s [=FileSystemHandle/entry=].
   1. If |options|.{{FileSystemGetFileOptions/create}} is `true`:
-    1. Run <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=request permission steps=] given
+    1. Run |entry|'s [=request permission steps=] given
        «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `true`&nbsp;]».
        If that throws an exception, [=reject=] |result| with that exception and abort.
     1. Let |permissionStatus| be the result of running
-       <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=query permission steps=] given
+       |entry|'s [=query permission steps=] given
        «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `true`&nbsp;]».
   1. Otherwise:
     1. Let |permissionStatus| be the result of running
-       <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=query permission steps=] given
+       |entry|'s [=query permission steps=] given
        «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `false`&nbsp;]».
   1. If |permissionStatus| is not {{PermissionState/"granted"}},
      reject |result| with a {{NotAllowedError}} and abort.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -96,6 +96,40 @@ run the following steps:
...
+access control for those entries.
+
+Issue: Figure out some way to integrate this better with the permissions API [[permissions]],
+for example by making the entry be a part of the PermissionDescriptor.
+
+<div algorithm>
+An [=/entry=] |entry| has associated <dfn for=entry>query permission steps</dfn>,
+which take a {{FileSystemHandlePermissionDescriptor}} and return a {{PermissionState}}.
+Unless specified otherwise, these steps are:
+1. Let |parent| be a [=directory entry=] where |parent|.[=children=] [=list/contains=] |entry|,
```
inexorabletash: This is a little awkward. 

How about defining "an entry's parent" somewhere? 

Also, is "a directory entry" intentionally indefinite? i.e. can there be more than one directory that contains the same entry?

mkruisselbrink: Thanks for reviewing this! I agree with it being awkward. I rephrased things a bit as you suggested, hopefully being clearer?

index.bs
```diff
@@ -396,7 +466,23 @@ iterable itself?
...
+1. Let |result| be [=a new promise=].
+1. Run the following steps [=in parallel=]:
+  1. If |options|.{{FileSystemGetFileOptions/create}} is `true`:
+    1. Run <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=request permission steps=] given
+       «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `true`&nbsp;]».
+       If that throws an exception, [=reject=] |result| with that exception and abort.
+    1. Let |permissionStatus| be the result of running
+       <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=query permission steps=] given
+       «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `true`&nbsp;]».
+  1. Else:
```
inexorabletash: "Otherwise" seems to be more common than "else"

index.bs
```diff
@@ -396,7 +466,23 @@ iterable itself?
...
+    1. Let |permissionStatus| be the result of running
+       <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=query permission steps=] given
+       «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `false`&nbsp;]».
+  1. If |permissionStatus| is not {{PermissionState/"granted"}},
+     reject |result| with a {{NotAllowedError}} and abort these steps.
```
inexorabletash: Consistency: "abort" vs. "abort these steps"

index.bs
```diff
@@ -242,7 +288,13 @@ The <dfn method for=FileSystemHandle>queryPermission(|descriptor|)</dfn> method,
 <div algorithm>
 The <dfn method for=FileSystemHandle>requestPermission(|descriptor|)</dfn> method, when invoked, must run these steps:
 
-1. TODO
+1. Let |result| be [=a new promise=].
+1. Run the following steps [=in parallel=]:
+  1. Run <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=query permission steps=] given |descriptor|.
+  1. If that throws an exception, [=reject=] |result| with that exception and abort.
```
inexorabletash: "that" is potentially confusing. Maybe "If the previous step throws an exception"

But see below...

index.bs
```diff
@@ -315,12 +375,22 @@ modifications to existing large files.
...
-1. Let |stream| be the result of [=create a new FileSystemWritableFileStream|creating a new FileSystemWritableFileStream=]
-   for <b>[=this=]</b>'s [=FileSystemHandle/entry=].
-1. If |options|.{{FileSystemCreateWritableOptions/keepExistingData}} is `true`:
-   1. Set |stream|.[=[[buffer]]=] to a copy of <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=file entry/binary data=].
-1. Return [=a promise resolved with=] |stream|.
+1. Let |result| be [=a new promise=].
+1. Run the following steps [=in parallel=]:
+  1. Run <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=request permission steps=] given
+     «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `true`&nbsp;]».
+     If that throws an exception, [=reject=] |result| with that exception and abort.
```
inexorabletash: Be consistent with the above, i.e. if the "catch" is part of the same step or not. Although it makes this line longer (and multi-part), I think I prefer this form, so maybe convert the above to be a single line? 

(No strong feelings here)

index.bs
```diff
@@ -96,6 +106,38 @@ run the following steps:
...
+1. If |parent| is null, return {{PermissionState/"granted"}}.
+1. Return the result of running |parent|'s [=query permission steps=].
+
+Note: These steps return the current state synchronously, however these steps are only
+invoked from "in parallel" sections of algorithms, so this doesn't have to be implemented
+in a synchronous manner.
+</div>
+
+<div algorithm>
+An [=/entry=] |entry| also has associated <dfn for=entry>request permission steps</dfn>,
```
inexorabletash: Maybe note that there's no return value, but the permission state may be updated and can be queried.

index.bs
```diff
@@ -594,13 +716,21 @@ To <dfn>create a new FileSystemWritableFileStream</dfn> given a [=file entry=] `
...
+      1. Let |permissionStatus| be the result of running
+         |stream|.[=[[file]]=]'s [=query permission steps=] given
+         «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `true`&nbsp;]».
+      1. If |permissionStatus| is not {{PermissionState/"granted"}},
+         reject |closeResult| with a {{NotAllowedError}} and abort these steps.
+      1. TODO: optional UA defined security checks
+      1. Set |stream|.[=[[file]]=]'s [=file entry/binary data=] to |stream|.[=[[buffer]]=].
+         If that throws an exception, reject |closeResult| with that exception and abort.
+
+         Note: it is expected that this atomically updates the contents of the file on disk
```
inexorabletash: Nit: Capitalize "It"

index.bs
```diff
@@ -783,6 +917,63 @@ steps:
...
+1. If the [=current global object=] is not a {{Window}},
+   return.
+
+   Issue: Should this throw instead of return silently?
+
+1. If the [=current global object=] does not have [=transient activation=], throw a {{NotAllowedError}}.
+1. Let |environment| be the [=current settings object=].
+1. If |environment|'s [=environment settings object/origin=] is not [=same origin=] with |environment|'s [=top-level origin=],
+   throw a {{NotAllowedError}}.
+1. Display a prompt to the user requesting access as described by |descriptor| to |entry|.
```
inexorabletash: I suspect this could be even more abstract, i.e. allowing a UA that never shows a prompt but always rejects, or infers an answer through other means.

I'd follow whatever other specs do. (I'm too lazy to check...)

index.bs
inexorabletash: Yep! And sorry for the single comment, I meant to batch up more...

index.bs

index.bs
mkruisselbrink: This sort of mostly matches what the permissions API uses for its request permission algorithm. I think the "never show a prompt" case is largely captured by the query permission steps. I.e. a UA can use whatever it wants to figure out what "most accurately reflects the user's intent", and if what most accurately reflects is in fact "prompt", this algorithm probably should in fact display a prompt (and if not, we won't even get to this step, since we'll just early abort this algorithm).

index.bs
mkruisselbrink: changed to "abort" throughout.

index.bs
```diff
@@ -81,7 +92,7 @@ run the following steps:
 1. Let |result| be [=a new promise=].
 1. Run the following steps [=in parallel=]:
   1. If |child| is [=the same as=] |root|,
-     [=/resolve=] |result| with an empty list, and abort these substeps.
+     [=/resolve=] |result| with an empty list, and abort.
```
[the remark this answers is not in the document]
inexorabletash: This is correct current spec-ese, although the language is constantly evolving. I notice Fetch is now using "abort these in-parallel steps" to be super clear. That may be overkill.

"Abort" isn't yet defined in https://infra.spec.whatwg.org/#algorithm-control-flow  but infra is a best effort attempt to document/define the language as it gets more rigorous.

index.bs
```diff
@@ -315,12 +379,22 @@ modifications to existing large files.
...
 
-1. Issue: TODO: Check permissions.
-1. Let |stream| be the result of [=create a new FileSystemWritableFileStream|creating a new FileSystemWritableFileStream=]
-   for <b>[=this=]</b>'s [=FileSystemHandle/entry=].
-1. If |options|.{{FileSystemCreateWritableOptions/keepExistingData}} is `true`:
-   1. Set |stream|.[=[[buffer]]=] to a copy of <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=file entry/binary data=].
-1. Return [=a promise resolved with=] |stream|.
+1. Let |result| be [=a new promise=].
+1. Run the following steps [=in parallel=]:
+  1. Run <b>[=this=]</b>'s [=FileSystemHandle/entry=]'s [=request permission steps=] given
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -783,6 +921,63 @@ steps:
...
+<div algorithm>
+The <dfn>native file system request permission steps</dfn> for an [=/entry=] |entry|
+given a {{FileSystemHandlePermissionDescriptor}} |descriptor| are:
+
+1. Let |status| be the result of running the [=query permission steps=] for |entry| given |descriptor|.
+1. If |status| is not {{PermissionState/"prompt"}} return.
+1. If the [=current global object=] is not a {{Window}},
+   return.
+
+   Issue: Should this throw instead of return silently?
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -783,6 +921,63 @@ steps:
...
+1. Display a prompt to the user requesting access as described by |descriptor| to |entry|.
+1. Wait for the user to accept or reject the prompt.
+   The user's interaction may provide [=new information about the user's intent=]
+   for this |descriptor| and |entry|.
+   Additionally if |descriptor|.{{FileSystemHandlePermissionDescriptor/writable}} is true,
+   the user's interaction may also provide [=new information about the user's intent=] for
+   the same |entry| with «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `false`&nbsp;]».
+
+Note: This is intentionally vague about the details of the permission UI and how the UA infers user intent.
+UAs <span class=allow-2119>should</span> be able to explore lots of UI within this framework.
```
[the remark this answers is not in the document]
mkruisselbrink: Done (this note was copied verbatim from the permission spec...)

index.bs
```diff
@@ -96,6 +107,41 @@ run the following steps:
 
 </div>
 
+## Permissions ## {#permissions}
+
+Note: By default entries inherit their permissions from their parent. If an entry doesn't
+have a parent, permissions are by default granted, and requesting permissions is a no-op.
```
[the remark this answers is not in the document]
mkruisselbrink: I rephrased things a bit, in particularly by getting rid of the "default granted" behavior. Does it seem clearer now?

index.bs
```diff
@@ -96,6 +107,41 @@ run the following steps:
...
+
+Issue: Figure out some way to integrate this better with the permissions API [[permissions]],
+for example by making the entry be a part of the PermissionDescriptor.
+
+<div algorithm>
+An [=/entry=] |entry| has associated <dfn for=entry>query permission steps</dfn>,
+which take a {{FileSystemHandlePermissionDescriptor}} and return a {{PermissionState}}.
+Unless specified otherwise, these steps are:
+1. Let |parent| be |entry|'s [=entry/parent=].
+1. If |parent| is null, return {{PermissionState/"granted"}}.
```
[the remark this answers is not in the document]
mkruisselbrink: Instead I now assert that |parent| is not null, and have the steps for the sandboxed file system defined by the section on the sandboxed file system.

PR #171, opened 2020-04-15
Be a bit more explciit about using the correct realm/settings object.

mkruisselbrink:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/171.html" title="Last updated on Apr 16, 2020, 8:59 PM UTC (a94412f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/171/b5e23db...a94412f.html" title="Last updated on Apr 16, 2020, 8:59 PM UTC (a94412f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -129,7 +129,7 @@ for example by making the entry be a part of the PermissionDescriptor.
 <div algorithm>
 An [=/entry=] |entry| has associated <dfn for=entry>query permission steps</dfn>,
 which take a {{FileSystemHandlePermissionDescriptor}} |descriptor|
-and an optional [=environment settings object=] |settings|,
+and an [=environment settings object=] |settings|,
 and return a {{PermissionState}}.
 Unless specified otherwise, these steps are:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -126,7 +128,9 @@ for example by making the entry be a part of the PermissionDescriptor.
 
 <div algorithm>
 An [=/entry=] |entry| has associated <dfn for=entry>query permission steps</dfn>,
-which take a {{FileSystemHandlePermissionDescriptor}} and return a {{PermissionState}}.
+which take a {{FileSystemHandlePermissionDescriptor}} |descriptor|
+and an optional [=environment settings object=] |settings|,
```
inexorabletash, line 132: Why optional? It seems this is passed in every invocation.

(here and the other "XXX permission steps", and therefore the use of "current settings object" if not passed)

mkruisselbrink: Optional because similar algorithms in the permissions spec had it optional. But agreed that there is no reason to have it optional here, so changed that.

PR #172, opened 2020-04-15
Be a bit more explicit about restricted directories.

### LINKED ISSUE #74: Restricted access to scary paths should be normative
In response to: https://github.com/w3ctag/design-reviews/issues/390

https://github.com/WICG/native-file-system/blob/master/security-privacy-questionnaire.md#27-does-this-specification-allow-an-origin-access-to-sensors-on-a-users-device

Suggests "no, /dev is off limits", but it feels like we should have some normative text to align behavior across limitations.

(e.g. "Please select your home directory", which a lot of Mac App Store apps do is even scary territory)

mkruisselbrink:
We don't want to dictate an explicit list of paths that all browsers
should block, but giving some examples of paths that might make sense
for user agents to block still seems sensible.

This fixes #74

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/172.html" title="Last updated on Apr 16, 2020, 8:55 PM UTC (2015779)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/172/b5e23db...2015779.html" title="Last updated on Apr 16, 2020, 8:55 PM UTC (2015779)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1375,9 +1375,11 @@ Examples of directories that user agents might want to restrict as being
 * Directories where the user agent stores [=site storage=].
 * Directories containing system files (such as `C:\Windows` on Windows).
 * Directories such as `/dev/`, `/sys`, and `/proc` on Linux that would give access to low-level devices.
-* A users entire "home" directory.
+* A user's entire "home" directory.
   Individual files and directories inside the home directory should still be allowed,
   but user agents should not generally let users give blanket access to the entire directory.
+* The default directory for downloads, if the user agent has such a thing.
+  Individual files inside the directory again should be allowed, but the whole directory would risk leaking more data than a user realizes.
 
 ## Websites trying to use this API for tracking. ## {#privacy-tracking}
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1368,6 +1368,17 @@ must be taken to strike the right balance between restricting what the API can a
...
 with some of their most private personal data.
 
+Examples of directories that user agents might want to restrict as being
+<dfn>too sensitive or dangerous</dfn> include:
+
+* The directory or directories containing the user agent itself.
+* Directories where the user agent stores [=site storage=].
+* Directories containing system files (such as `C:\Windows` on Windows).
+* Directories such as `/dev/`, `/sys`, and `/proc` on Linux that would give access to low-level devices.
+* A users entire "home" directory.
```
pwnall, line 1378: users -> user's

index.bs
```diff
@@ -1368,6 +1368,17 @@ must be taken to strike the right balance between restricting what the API can a
...
+  Individual files and directories inside the home directory should still be allowed,
+  but user agents should not generally let users give blanket access to the entire directory.
+
```
pwnall, line 1381: WDYT about mentioning the default directory for downloads? (`<a download>` and downloads with `Content-Disposition: attachment`)

Issue #27, opened 2018-08-31
Bikeshedding getSystemDirectory name


inexorabletash:
`getSystemDirectory()` could be used for things like `"sandbox"` (see #10) so "system" might not be the best term for the method name

* Windows calls these "Known Folders" [link](https://docs.microsoft.com/en-us/windows/desktop/shell/working-with-known-folders)
* Android calls these "Standard Directories" [link](https://developer.android.com/reference/android/os/Environment)
* MacOS docs appear to have enum values but not a common term for these. 

Ideas:

* getStandardDirectory
* getNamedDirectory
* getSpecialDirectory
* getKnownDirectory
* getSystemDirectory

... or change the API shape and make there be an enum that you pass in to `choose()` with the behavior that you get a chooser if the system doesn't support such a type?

mkruisselbrink:
I think it might make sense to separate sandboxed file system access from known/standard directory access. At the time it seemed nice to combine these as one, but after thinking about it more I think it might just be confusing.

So I'd propose replacing the current `getSystemDirectory({type: "sandbox"})` with simply `getSandboxedFileSystem()` (or whatever other name makes sense), and in the future when we introduce some kind of special directory access we can figure out the best shape of such an API separately.

ftreesmilo:
Can this be closed?

mkruisselbrink:
Yes, I think so. We have no current plans to do any work on features like this.

PR #174, opened 2020-04-16
Change the API for exposing the sandboxed file system.

### LINKED ISSUE #66: Explain what sandboxed file system is.
http://wicg.github.io/native-file-system/#api-getsystemdirectory mentions it returning "the sandboxed file system" without any indication what that actually is. Other than actually formally specifying what this is about, we should at least give some explanation in spec and explainer what this is and why we want to expose it.

(fwiw, it is intended to be more or less the same as the "temporary" filesystem from https://dev.w3.org/2009/dap/file-system/file-dir-sys.html#temporary-vs-persistent-storage. I.e. a quota-managed, origin-scoped private storage area)

mkruisselbrink:
Also explain what and why the sandboxed file system is.

This fixes #66 and partially addresses #27.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/174.html" title="Last updated on Apr 17, 2020, 11:37 PM UTC (29a6465)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/174/827c379...29a6465.html" title="Last updated on Apr 17, 2020, 11:37 PM UTC (29a6465)">Diff</a>

### DIFF
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -28,6 +28,8 @@ In native applications, there are common file access patterns that we aim to add
 1. The files can be opened in any native or web applications concurrently
 1. Changes to the files on disk, made in any native or other web application, are accessible
 1. Access the files with the same access in future browsing sessions
+1. Create a new file in the editor
+1. Auto-save changes to the new file in a temporary location, even before the user has picked a file name/location
 
 ### Multi-file Editor
 1. Open a directory that contains many files and sub-directories, represented hierarchically
@@ -73,6 +75,13 @@ Some example applications of the API we would like to address:
 But even though we'd like to design the API to eventually enable all these use
 cases, initially we'd almost certainly be shipping a very limited API surface
 with limited capabilities.
+
+We also want to make it possible for websites to get access to some directory
+without having to first prompt the user for access. This helps to make
+it easier to write automated tests. It also enables use cases where a website
+wants to save data to disk before a user has picked a location to save to,
+without forcing the website to use a completely different storage mechanism
+with a different API for such files.
 
 ## Non-goals
 
@@ -189,7 +198,7 @@ request.onsuccess = function(e) {
 
     // Rejects if file is no longer writable, because the website no longer has
     // permission to write to it.
-    let file_writer = await ref.createWritable({createIfNotExists: true});
+    let file_writer = await ref.createWritable();
     // ... write to file_writer
 }
 ```
@@ -230,7 +239,7 @@ const file_ref = await dir_ref.getFile('foo.js');
 // Do something useful with the file.
 
 // Get a subdirectory.
-const subdir = await dir_ref.getDirectory('bla', {createIfNotExists: true});
+const subdir = await dir_ref.getDirectory('bla', {create: true});
 
 // No special API to create copies, but still possible to do so by using
 // available read and write APIs.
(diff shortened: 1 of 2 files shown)


mkruisselbrink:
@inexorabletash @pwnall what do you think of this change? Decoupling sandboxed file system from "special/system/standard/well known" directories seems like it would be less confusing to me...

EXPLAINER.md
```diff
@@ -74,6 +76,13 @@ But even though we'd like to design the API to eventually enable all these use
 cases, initially we'd almost certainly be shipping a very limited API surface
 with limited capabilities.
 
+We also want to make it possible for websites to get access to some directory
+without having to first prompt the user for access. This helps to make
```
inexorabletash, line 80: I'd order these examples by "hierarchy of constituencies", so put the "write automated tests" example last.

Do we also want to call out the generic use case of storing internal application state in mutable file-like data where neither Indexed DB nor Cache Storage are good fits?

index.bs
```diff
@@ -1324,43 +1332,31 @@ The [=sandboxed file system root=]'s [=query permission steps=] are the followin
...
-
-dictionary GetSystemDirectoryOptions {
-  required SystemDirectoryType type;
-};
-
 [SecureContext]
-partial interface FileSystemDirectoryHandle {
-  static Promise<FileSystemDirectoryHandle> getSystemDirectory(GetSystemDirectoryOptions options);
+partial interface Window {
+  Promise<FileSystemDirectoryHandle> getSandboxedFileSystem();
```
inexorabletash, line 1340: Bikeshedding on the name a bit... 

* This is the only place "FileSystem" appears in the JS API, right? (not counting interface names). Should it just be "Directory" instead?  Behind the scenes we implement it as a separate namespace, but... from the perspective of this API which doesn't provide the same FS root like the legacy API, this is just a directory.

* Maybe "origin" instead of "sandboxed" so we're not throwing a new term into JS? Not sure about this... "getOriginDirectory()" seems a overbroad. "getPrivateDirectory()" ? "getOriginPrivateDirectory()" ? 

Sorry for not including this bikeshedding in the issue; I meant to LGTM the idea of a separate method, but discuss the name further. (no objections to landing the PR with the current name, though!)

index.bs
```diff
@@ -1324,43 +1332,31 @@ The [=sandboxed file system root=]'s [=query permission steps=] are the followin
...
 };
 </xmp>
 
+Advisement: In Chrome this functionality is currently exposed as `FileSystemDirectoryHandle.getSystemDirectory({type: "sandbox"})`.
```
inexorabletash, line 1344: Add the milestone? I also like tossing in 🚧 markers, but I'm silly that way.

EXPLAINER.md
mkruisselbrink: swapped the order. Not sure about such generic use cases... Some more specific use case might make sense for that, but I'm not sure if just describing the capability makes for a very compelling use case.

index.bs
mkruisselbrink: Well, currently it's all milestones... Once we've implemented these changes I definitely want to go back and update this to mention milestones, just not sure what milestone to mention now.

index.bs
mkruisselbrink: I like getOriginPrivateDirectory, so changed it to that for now.

PR #175, opened 2020-04-17
Call out malware scans in writer close algorithm.

### LINKED ISSUE #37: Potentially Unwanted Programs Scan

mkruisselbrink:
This makes it clear when user agents can perform such scans, and how
failing a check will be reported.

Fixes #37

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/175.html" title="Last updated on May 5, 2020, 5:07 PM UTC (5706665)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/175/de41e45...5706665.html" title="Last updated on May 5, 2020, 5:07 PM UTC (5706665)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -844,7 +844,7 @@ in a [=/Realm=] |realm|, perform the following steps:
          and |stream|'s [=relevant settings object=].
       1. If |permissionStatus| is not {{PermissionState/"granted"}},
          reject |closeResult| with a {{NotAllowedError}} and abort.
-      1. Perform any [=malware scans and safe browsing checks=] the user agent might want to do.
+      1. Perform user agent-specific [=malware scans and safe browsing checks=].
          If these checks fail, [=/reject=] |closeResult| with an {{AbortError}} and abort.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -836,9 +836,10 @@ in a [=/Realm=] |realm|, perform the following steps:
          and |stream|'s [=relevant settings object=].
       1. If |permissionStatus| is not {{PermissionState/"granted"}},
          reject |closeResult| with a {{NotAllowedError}} and abort.
-      1. TODO: optional UA defined security checks
+      1. Perform any [=malware scans and safe browsing checks=] the user agent might want to do.
```
pwnall, line 839: nit: any [=malware scans and safe browsing checks=] the user agent might want to do -> user agent-specific [=malware scans and safe browsing checks=]?

(Assuming this doesn't lose precision, I think it's a bit shorter.)

PR #177, opened 2020-05-04
Properly specify async iterable behavior.

### LINKED ISSUE #47: Async iterables are happening
@Ms2ger is working on Web IDL support over in https://github.com/heycam/webidl/pull/720. Your review would be appreciated, and then once it's ready, you can use it.

### LINKED ISSUE #127: getEntries() iteration order in the face of changes?
https://wicg.github.io/native-file-system/#api-filesystemdirectoryhandle-getentries is a TODO for normative text, and one thing that seems important to define well is what happens if the entries change during iteration. It's probably a good idea to define the iteration order even for a static case to not depend on the underlying file system's order, but then entries added during iteration would have to either not show up at all, or show up at the end out-of-order.

This would affect https://github.com/WICG/native-file-system/issues/47 in the same way, cc @domenic @Ms2ger.

### LINKED ISSUE #158: Symbol.asyncIterator on FileSystemDirectoryHandle?
have experimented with the API a bit, could not help but feel like `FileSystemDirectoryHandle` is missing `Symbol.asyncIterator` that maps to `getEntries`

```js
for await (const entry of dirHandle) { ... }
```

mkruisselbrink:
This fixes #158, fixes #127, and fixes #47.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/177.html" title="Last updated on May 4, 2020, 7:11 PM UTC (d815ef1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/177/de41e45...d815ef1.html" title="Last updated on May 4, 2020, 7:11 PM UTC (d815ef1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -460,11 +460,10 @@ dictionary FileSystemRemoveOptions {
 
 [Exposed=(Window,Worker), SecureContext, Serializable]
 interface FileSystemDirectoryHandle : FileSystemHandle {
+  async iterable<USVString, FileSystemHandle>;
+
   Promise<FileSystemFileHandle> getFile(USVString name, optional FileSystemGetFileOptions options = {});
   Promise<FileSystemDirectoryHandle> getDirectory(USVString name, optional FileSystemGetDirectoryOptions options = {});
-
-  // This really returns an async iterable, but that is not yet expressable in WebIDL.
-  object getEntries();
 
   Promise<void> removeEntry(USVString name, optional FileSystemRemoveOptions options = {});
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -487,8 +486,82 @@ Issue(98): Having getFile methods in both FileSystemDirectoryHandle and FileSyst
...
+  : for await (let [|name|, |handle|] of |directoryHandle|) {}
+  : for await (let [|name|, |handle|] of |directoryHandle| . entries()) {}
+  : for await (let |handle| of |directoryHandle| . values()) {}
+  : for await (let |name| of |directoryHandle| . keys()) {}
+  :: Iterates over all entries whose parent is the entry represented by |directoryHandle|. Entries
+     that are created or deleted while the iteration is in progress might or might not be included.
+     No guarantees are given either way.
+</div>
+
+Advisement: In Chrome this is currently implemented as a `directoryHandle.getEntries()` method that can be used in a `for await..of` loop.
```
pwnall, line 501: Would we fix this in Chrome 84? If so, should we call this out?

mkruisselbrink: We should definitely update this to mention the milestone chrome implements this, when we know what milestone that is. Not sure currently if that is going to 84 branching next week, but I suspect not. So leaving it like this for now.

PR #185, opened 2020-06-10
Change the API surface for getting native file system handles.

### LINKED ISSUE #25: Should `choose` always return an array?
In the current strawman proposal the `choose` style API returns a single handle if `multiple` is false (or not specified), and returns a sequence if `multiple` is true (even if in that case only a single file was selected). Would it be more/less confusing if the API always returned an array?

### LINKED ISSUE #133: How do MIME types map to native file systems?
[`chooseFileSystemEntries()`](https://wicg.github.io/native-file-system/#dom-window-choosefilesystementries) supports MIME types and file extensions. The relevant part of the algorithm is still a TODO, but spelling out how one uses a MIME type to filter a list of candidate entries seems very important. It's probably one of:
- Mapping a MIME type to a set of extensions
- Sniffing the file content
- Some file system or OS-level feature?

Any mismatch here could leave a site broken and unusable in some browsers, so this seems to warrant a fair bit of detail.

### LINKED ISSUE #134: What does excludeAcceptAllOption do?
For [`chooseFileSystemEntries()`](https://wicg.github.io/native-file-system/#dom-window-choosefilesystementries) has an `excludeAcceptAllOption` options, non-normatively described as:
> If options.excludeAcceptAllOption is true, or if no valid entries exist in options.accepts, a option matching all files will be included in the file types the dialog lets the user select.

Given the name and this description, I'm still not sure what it does. From the name it sounds like maybe it removes some checkbox or dropdown option from the native OS file picker? But from the description it sounds more like if there are no matches, then this will cause all files to become eligible for selection instead?

### LINKED ISSUE #170: mimeTypes or just types ?
In https://wicg.github.io/native-file-system/#dictdef-choosefilesystementriesoptionsaccepts

The FileAPI just uses 'type'. While a few places in the web platform (webrtc) use "mimeType", I think we're trying to avoid using "mime" these days. Should this dictionary member just be "types" ?

mkruisselbrink:
Rather than having one method do three different things, just have
separate methods for all three options. Also aligns (mostly) with
the file handling API in how accepted file types are specified.

This fixes #170, fixes #134, fixes #133 and fixes #25.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/185.html" title="Last updated on Jun 11, 2020, 12:00 AM UTC (0677301)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/185/9f209a4...0677301.html" title="Last updated on Jun 11, 2020, 12:00 AM UTC (0677301)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -91,7 +91,7 @@ An [=/entry=]'s [=entry/parent=] is null if no such directory entry exists.
 Note: Two different [=/entries=] can represent the same file or directory on disk, in which
 case it is possible for both entries to have a different parent, or for one entry to have a
 parent while the other entry does not have a parent. Typically an entry does not have a parent
-if it was returned by {{getOriginPrivateDirectory()}} or one of the [=native file system entry points=],
+if it was returned by {{getOriginPrivateDirectory()}} or one of the [=native file system handle factories=],
 and an entry will have a parent in all other cases.
 
 [=/Entries=] can (but don't have to) be backed by files on the systems native file system,
@@ -144,7 +144,7 @@ Unless specified otherwise, these steps are:
 1. Assert: |parent| is not null.
 
    Note: [[#native-file-system-permissions]] overrides these steps for entries returned by
-   the [=native file system entry points=]. Additionally [[#sandboxed-filesystem-concepts]] overrides
+   the [=native file system handle factories=]. Additionally [[#sandboxed-filesystem-concepts]] overrides
    these steps for entries returned by {{getOriginPrivateDirectory()}}.
    All other entries always have a parent.
 
@@ -273,7 +273,7 @@ Issue(119): the currently described API here assumes a model where it is not pos
      the website will have to call {{FileSystemHandle/requestPermission()}} before any
      operations on the handle can be done. If this returns `"denied"` any operations will reject.
 
-     Usually handles returned by the [=native file system entry points=] will initially return `"granted"` for
+     Usually handles returned by the [=native file system handle factories=] will initially return `"granted"` for
      their read permission state, however other than through the user revoking permission, a handle
      retrieved from IndexedDB is also likely to return `"prompt"`.
 
@@ -1116,7 +1116,7 @@ partial interface Window {
 </xmp>
 
 The {{showOpenFilePicker()}}, {{showSaveFilePicker()}} and {{showDirectoryPicker()}} methods
-are together known as the <dfn>native file system entry points</dfn>.
+are together known as the <dfn>native file system handle factories</dfn>.
 
 ## Native File System Permissions ## {#native-file-system-permissions}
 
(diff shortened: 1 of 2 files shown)


mkruisselbrink:
Still need to update the explainer, add warnings that chrome doesn't implement this yet, etc. But feedback/bikeshedding on method names, exact API shape etc is very welcome.

Should also probably resolve https://github.com/WICG/file-handling/issues/43 to truly get consistence between this and the file handling API.

index.bs
```diff
@@ -1327,14 +1264,231 @@ these steps:
...
+
+<div class="note domintro">
+The {{showFilePicker(options)}} and {{showSaveFilePicker(options)}} methods accept a
+{{FilePickerOptions}} argument, which lets the website specify the types of files
+the file picker will let the user select.
+
+Each entry in {{FilePickerOptions/types}} specifies a single user selectable option
+for filtering the files displayed in the file picker.
+
+Each option consists of an <span class=allow-2119>optional</span> {{FilePickerAcceptType/description}}
```
pwnall, line 1354: I think we're introducing "description" for the first time. Two concerns that can be addressed in follow-ups:

1) Description is the first user-visible (I assume?) bit of info in this section. Do we need a section on internalization concerns now?
2) Do we want non-normative advice on how the description would be used in the file picker? At least in Chrome, I think we try to avoid showing app-supplied strings in system UI (like file pickers).

tomayac:
Coming late to the party—and apologies if this was discussed before—but what about renaming `showFilePicker()` to `showOpenFilePicker()` to be aligned with the others?

mkruisselbrink:
> Coming late to the party—and apologies if this was discussed before—but what about renaming `showFilePicker()` to `showOpenFilePicker()` to be aligned with the others?

Not late at all. I like that suggestion, will do.

index.bs
mkruisselbrink: The old API had the same, but yeah, we definitely should have such a section. Filed #186 and #187.

Good point about displaying app-supplied strings in system UI. We've been doing that all along, and nobody has complained yet, but would probably be good to say something about in the spec. Filed #188 to track that.

index.bs
```diff
@@ -91,8 +91,8 @@ An [=/entry=]'s [=entry/parent=] is null if no such directory entry exists.
 Note: Two different [=/entries=] can represent the same file or directory on disk, in which
 case it is possible for both entries to have a different parent, or for one entry to have a
 parent while the other entry does not have a parent. Typically an entry does not have a parent
-if it was returned by {{chooseFileSystemEntries()}} or {{getOriginPrivateDirectory()}}, and an entry
-will have a parent in all other cases.
+if it was returned by {{getOriginPrivateDirectory()}} or one of the [=native file system entry points=],
```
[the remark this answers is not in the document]
mkruisselbrink: sgtm, done

PR #192, opened 2020-06-23
Updated specs to include Drag and Drop section

### LINKED ISSUE #82: Integration with drag-and-drop files
It would be nice if the native filesystem API could cover dragging and dropping a file in to the web app, and still being able to write to that file.

Currently as per spec it appears you can only get write access to a file if you use `chooseFileSystemEntries` in a user gesture and the user selects the file in a file picker dialog. Our web app (and presumably many others) support opening files both by choosing a file (via `<input type="file">` or drag-and-drop (via the "drop" event). `chooseFileSystemEntries` only covers the former, not the latter. This creates a discrepancy in the user experience where the particular way they open a file ends up affecting whether or not the web app is allowed to save back over that file. If some users only ever use drag-and-drop to open files, we may even end up in the situation where we support the native file system API, but some users don't benefit because they open files a different way.

I'd have thought that the user dragging-and-dropping a file in to a web app equally indicates their intention for the web app to access it. Since writing back requires a prompt, there is still no automatic write ability, it would basically be an alternative to an open prompt with `chooseFileSystemEntries`.

It's not clear to me how this would be supported, since the "drop" event uses DataTransfer.files to provide the files, which are ordinary read-only File objects instead of writable FileSystemFileHandles. Perhaps it could either:

1. Add a separate DataTransfer.fileHandles list
2. Add a method to get a file handle back from a file

HazimMohamed:
I added a section to the docs to explain how Drag and Drop is going to be integrated into the Native File System API.

***
<a href="https://pr-preview.s3.amazonaws.com/HazimMohamed/native-file-system/pull/192.html" title="Last updated on Aug 14, 2020, 5:47 PM UTC (c8b65f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/192/81f49a8...HazimMohamed:c8b65f0.html" title="Last updated on Aug 14, 2020, 5:47 PM UTC (c8b65f0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1466,8 +1466,85 @@ these steps:
   1. [=/Resolve=] |p| with |result|.
 
 1. Return |p|.
+</div>
+
+## Drag and Drop ## {#drag-and-drop}
+
+<xmp class=idl>
+partial interface DataTransferItem {
+    Promise<FileSystemHandle?> getAsFileSystemHandle();
+};
+</xmp>
+
+During a <em>drag-and-drop operation</em>, dragged file and
+directory items are associated with [=file entries=] and [=directory entries=]
+respectively.
+
+<div class="note domintro">
+  : |handle| = await item . {{getAsFileSystemHandle()}}
+  :: Returns a {{FileSystemFileHandle}} object if the dragged item is file and a {{FileSystemDirectoryHandle}} object if the dragged item is a directory.
+</div>
+
+<div algorithm>
+
+The <dfn method for=DataTransferItem>getAsFileSystemHandle()</dfn> method steps are:
+
+1. Let |p| be [=a new promise=].
+
+1. Run the following steps [=in parallel=]:
+
+  1. If the {{DataTransferItem}} object is not in the <a spec=html>read/write
+      mode</a> or the <a spec=html>read-only mode</a>, then [=/resolve=] |p| with `null`.
+
+  1. If the <a spec=html>the drag data item kind</a> is not <em>File</em>, then 
+      then [=/resolve=] |p| with `null`.
+
+  1. Let |entry| be the [=/entry=] representing the dragged file or directory.
+
+  1. If |entry| is a [=file entry=]:
+
+    1. Let |handle| be a {{FileSystemFileHandle}} associated with |entry|.
+
+  1. Else if |entry| is a [=directory entry=]:
+
+    1. Let |handle| be a {{FileSystemDirectoryHandle}} associated with |entry|.
+
+  1. [=/Resolve=] |p| with |entry|.
+
+1. Return |p|.
 
 </div>
+
+<div class=example id=draganddrop-example>
+Handling drag and drop of files and directories:
+<xmp highlight=js>
+elem.addEventListener('dragover', (e) => {
+  // Prevent navigation.
+  e.preventDefault();
+});
+elem.addEventListener('drop', async (e) => {
+  // Prevent navigation.
+  e.preventDefault();
+
+  // Process all of the items.
+  for (const item of e.dataTransfer.items) {
+    // kind will be 'file' for file/directory entries.
+    if (item.kind === 'file') {
+      const entry = await item.getAsFileSystemHandle();
+      if (entry.kind == 'file') {
+        handleDirectoryEntry(entry);
+      } else if (entry.kind == 'directory') {
+        handleFileEntry(entry);
+      }
+    }
+  }
+});
+</xmp>
+</div>
+
+Issue: This currently does not block access to [=too sensitive or dangerous=] directories, to
+be consistent with other APIs that give access to dropped files and directories. This is inconsistent
+with the [=native file system handle factories=] though, so we might want to reconsider this.
 
 # Accessing the Origin Private File System # {#sandboxed-filesystem}
 
--- a/index.bs
+++ b/index.bs
@@ -1466,6 +1466,7 @@ these steps:
   1. [=/Resolve=] |p| with |result|.
 
 1. Return |p|.
+
 </div>
 
 ## Drag and Drop ## {#drag-and-drop}
@@ -1489,15 +1490,16 @@ respectively.
 
 The <dfn method for=DataTransferItem>getAsFileSystemHandle()</dfn> method steps are:
 
+1. If the {{DataTransferItem}} object is not in the <a spec=html>read/write
+    mode</a> or the <a spec=html>read-only mode</a>, then return 
+    [=a promise resolved with=] `null`.
+
+1. If the <a spec=html>the drag data item kind</a> is not <em>File</em>, then 
+    then return [=a promise resolved with=] `null`.
+
 1. Let |p| be [=a new promise=].
 
 1. Run the following steps [=in parallel=]:
-
-  1. If the {{DataTransferItem}} object is not in the <a spec=html>read/write
-      mode</a> or the <a spec=html>read-only mode</a>, then [=/resolve=] |p| with `null`.
-
-  1. If the <a spec=html>the drag data item kind</a> is not <em>File</em>, then 
-      then [=/resolve=] |p| with `null`.
 
   1. Let |entry| be the [=/entry=] representing the dragged file or directory.
 
--- a/index.bs
+++ b/index.bs
@@ -1491,10 +1491,10 @@ respectively.
 The <dfn method for=DataTransferItem>getAsFileSystemHandle()</dfn> method steps are:
 
 1. If the {{DataTransferItem}} object is not in the <a spec=html>read/write
-    mode</a> or the <a spec=html>read-only mode</a>, then return 
+    mode</a> or the <a spec=html>read-only mode</a>, return 
     [=a promise resolved with=] `null`.
 
-1. If the <a spec=html>the drag data item kind</a> is not <em>File</em>, then 
+1. If the <a spec=html>the drag data item kind</a> is not <em>File</em>, 
     then return [=a promise resolved with=] `null`.
 
 1. Let |p| be [=a new promise=].
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -88,7 +88,7 @@ to write automated tests for code using this API.
 At least for now out of scope is access to the full file system, subscribing to
 file change notifications, probably many things related to file metadata (i.e.
 marking files as executable/hidden, etc). Also not yet planning to address how
-this new API might integrate with drag&drop and `<input type=file>`.
+this new API might integrate with `<input type=file>`.
 
 # Example code
 


marcoscaceres:
Thanks for your contribution, @HazimMohamed. You should have received an email from the W3C's IPR bot. When you can, please follow the instructions there for joining the WICG.

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
 
 </div>
 
+# Drag and Drop Integration # {#drag-and-drop}
+
+During a <em>drag-and-drop operation</em>, dragged file and
+directory items are associated with [=file entries=] and [=directory entries=] 
+respectively. Each [=/entry=] is stored as an element in the <l spec=html>{{DragEvent|DragEvent's}}</l>
+<l spec=html>[=drag data store=] </l> and can be accessed using {{DataTransferItem/getAsFileSystemEntry()}} on
```
mkruisselbrink, line 1573: nit: Just `<a spec=html>drag data store</a>`. The `<l>` bit can be useful for other link shorthands (i.e. `<l spec=html>{{Foo/bar()}}</l>` is shorthand for `<a method for=Foo spec=html>bar()</a>`), but there really isn't any benefit to it for regular "dfn" definitions.

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+the entry's associated <l spec=html>{{DataTransferItem}}</l>.
```
mkruisselbrink, line 1574: no need for the explicit spec

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+
+## The getAsFileSystemEntry() method ## {#api-getasfilesystementry}
```
mkruisselbrink, line 1576: nit: #api-datatransferitem-getasfilesystementry, or since this entire section is pretty small anyway, I don't think we even need this extra header

index.bs
```diff
@@ -145,7 +145,8 @@ Unless specified otherwise, these steps are:
 
    Note: [[#native-file-system-permissions]] overrides these steps for entries returned by
    the [=native file system handle factories=]. Additionally [[#sandboxed-filesystem-concepts]] overrides
-   these steps for entries returned by {{getOriginPrivateDirectory()}}.
+   these steps for entries returned by {{getOriginPrivateDirectory()}}, and the gesture of dragging
+   and dropping overrides these steps for entries dropped onto the web.
```
mkruisselbrink, line 149: (after making drag&drop be a subsection of accessing native file system), I'd rephrase this as 

[[#nfs-permissions]] overrides these steps for entries returned by factories, as well as entries exposed via drag and drop

With a link to the drag&drop section,

index.bs
```diff
@@ -1158,7 +1159,7 @@ resolves, the [=query permission steps=] for the returned entries
 given «[&nbsp;"{{FileSystemHandlePermissionDescriptor/writable}}" → `false`&nbsp;]»
 should return {{PermissionState/"granted"}}.
 
-Additionally for calls to {{showSaveFilePicker}
+Additionally for calls to {{showSaveFilePicker}}
```
mkruisselbrink, line 1162: Thanks for fixing this!

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+accessed via the [=native file system handle factories=], with the exception that [=too sensitive or dangerous|sensitive directories=]
+are not blocked via drag and drop.
+
+Issue: Discuss blocking when dragging forbidden sensitive directories onto the web. For
+now these directories can be accessed through drag and drop.
+
+Issue: Eventually there should be discussion about integrating drag and drop operations
+with `<input type='file'>` elements, but for now, there is no interaction.
+
+<xmp class=idl>
```
mkruisselbrink, line 1592: I'd move the IDL definition to the top of this sub-section,

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+};
+</xmp>
+
+<div algorithm>
+
+The <dfn method for=DataTransferItem>getAsFileSystemEntry()</dfn> method steps are:
+
+1. If the {{DataTransferItem}} object is not in the <l spec=html> <a>read/write
+    mode</a></l> or the <l spec=html>[=read-only mode=]</l>, return null and abort these
+    steps.
```
mkruisselbrink, line 1604: no need for the "abort these steps" bit. Per https://infra.spec.whatwg.org/#algorithm-control-flow, "a requirement to return terminates the algorithm the statement was in"

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+    FileSystemHandle? getAsFileSystemEntry();
+};
+</xmp>
+
+<div algorithm>
+
+The <dfn method for=DataTransferItem>getAsFileSystemEntry()</dfn> method steps are:
+
+1. If the {{DataTransferItem}} object is not in the <l spec=html> <a>read/write
+    mode</a></l> or the <l spec=html>[=read-only mode=]</l>, return null and abort these
```
mkruisselbrink, line 1603: here too, I would just write these references as `<a spec=html>read/write mode</a>` etc

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+    steps.
+
+1. If the <l spec=html>[=the drag data item kind=]</l> is not <em>File</em>, then return
+    null and abort these steps.
+
+1. If the dragged item is a file:
+  
+  1. Let |handle| be a {{FileSystemFileHandle}} representing the dragged file
```
mkruisselbrink, line 1611: in these steps you're confusing |handle|, the javascript object, and its associated entry a bit. I think I'd roughly write this as:
```
1. Let |entry| be the [/=entry=] representing the dragged file or directory.
1. Set |entry|'s permission steps, etc.
1. If |entry| is  [=file entry=]:
  1. Let |handle| be a .... associated with |entry|.
etc...
```

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+
+1. Set |handle|'s [=request permission steps=] to the [=native file system request permission steps=]
+
+1. Return <em>handle</em>
+
+</div>
+
+<div class=example id=draganddrop-example>
+<xmp highlight=js>
+Handling drag and drop of files and directories:
```
mkruisselbrink, line 1627: This line should probably be outside the syntax highlighted <xmp>.

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+
+Issue: Discuss blocking when dragging forbidden sensitive directories onto the web. For
+now these directories can be accessed through drag and drop.
+
+Issue: Eventually there should be discussion about integrating drag and drop operations
+with `<input type='file'>` elements, but for now, there is no interaction.
+
+<xmp class=idl>
+partial interface DataTransferItem {
+    FileSystemHandle? getAsFileSystemEntry();
```
mkruisselbrink, line 1594: It's kind of confusing how sometimes we're talking about handles and sometimes about entries, but this method should probably actually be getAsFileSystemHandle, given that that is the type that is being returned.

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
 
 </div>
 
+# Drag and Drop Integration # {#drag-and-drop}
+
+During a <em>drag-and-drop operation</em>, dragged file and
+directory items are associated with [=file entries=] and [=directory entries=] 
+respectively. Each [=/entry=] is stored as an element in the <l spec=html>{{DragEvent|DragEvent's}}</l>
```
mkruisselbrink, line 1572: I would actually just leave out the entire sentence starting with "Each entry is stored...". We could duplicate the `Additionally, each directory item is represented in the drag data store item list as a File. If it is accessed via getAsFile() a zero-length File is returned.` sentence from entries-API, as that is still true here as well, but it's probably fine to just leave that out as well.

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+<l spec=html>[=drag data store=] </l> and can be accessed using {{DataTransferItem/getAsFileSystemEntry()}} on
+the entry's associated <l spec=html>{{DataTransferItem}}</l>.
+
+## The getAsFileSystemEntry() method ## {#api-getasfilesystementry}
+
+{{DataTransferItem/getAsFileSystemEntry()}} will return a {{FileSystemFileHandle}} object if
```
mkruisselbrink, line 1578: I don't think this paragraph makes sense as a normative requirement (by default all text in a spec is considered normative). But this could be a "domintro" block, i.e. something like:
```
<div class="note domintro">
: |handle| = item . {{getAsFileSystemEntry()}}
:: Returns a ...
</div>
```

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+the dragged item is file and a {{FileSystemDirectoryHandle}} object if the dragged item
+is a directory.
+
+Dragging and dropping a file/directory will grant the same permissions as if they were
+accessed via the [=native file system handle factories=], with the exception that [=too sensitive or dangerous|sensitive directories=]
+are not blocked via drag and drop.
+
+Issue: Discuss blocking when dragging forbidden sensitive directories onto the web. For
+now these directories can be accessed through drag and drop.
```
mkruisselbrink, line 1587: I think I'd combine this with the previous paragraph (and perhaps move it till after the algorithm definition), something like:

Issue: This currently does not block access to [=too sensitive or dangerous=] directories, to be consistent with other APIs that give access to dropped files and directories. This is inconsistent with the [=nfs handle factories=] though, so we might want to reconsider this.

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
...
+
+Issue: Eventually there should be discussion about integrating drag and drop operations
```
mkruisselbrink, line 1589: I think I'd just leave this out for now.

mkruisselbrink (changes requested):
Thanks for writing this down. As Marcos pointed out you will need to join the W3C and WICG on behalf of Google before we can merge this.

I thing this largely looks okay. I think I left a bunch of confusing comments though. At a high level I think I'd structure this change by adding a "Drag and drop integration" subsection the the "Accessing Native File System" section. And within that section start out with your first paragraph, followed by the web-idl, followed by a domintro block, followed by the example (it seems currently I'm inconsistent about having the example inside a domintro block, or outside it; either is fine), followed by the algorithm, and finally the issue about what to do about sensitive directories.

Does that make sense?

mkruisselbrink:
Oh, one other comment, can you also update EXPLAINER.md to remove drag&drop from the non-goals?

index.bs
```diff
@@ -1564,6 +1565,90 @@ invoked, must run these steps:
 
 </div>
 
+# Drag and Drop Integration # {#drag-and-drop}
```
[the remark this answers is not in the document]
HazimMohamed: Done.

index.bs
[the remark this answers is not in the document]
HazimMohamed: Done.

index.bs
HazimMohamed: Done.

index.bs
HazimMohamed: Done.

index.bs
HazimMohamed: I just removed it 👍

index.bs
HazimMohamed: Done.

index.bs
HazimMohamed: I just removed it.

index.bs
```diff
@@ -144,9 +144,7 @@ Unless specified otherwise, these steps are:
 1. Assert: |parent| is not null.
 
    Note: [[#native-file-system-permissions]] overrides these steps for entries returned by
-   the [=native file system handle factories=]. Additionally [[#sandboxed-filesystem-concepts]] overrides
```
mkruisselbrink, line 147: Sorry if I was unclear. I do want to keep the rest of this paragraph as well.

index.bs
```diff
@@ -1507,6 +1505,81 @@ these steps:
...
+: |handle| = item . {{getAsFileSystemEntry()}}
+:: Returns a {{FileSystemFileHandle}} object if the dragged item is file and a {{FileSystemDirectoryHandle}} object if the dragged item is a directory.
+</div>
+
+<div algorithm>
+
+The <dfn method for=DataTransferItem>getAsFileSystemEntry()</dfn> method steps are:
+
+1. If the {{DataTransferItem}} object is not in the <a spec=html>read/write
+    mode</a> or the <a spec=html>[=read-only mode=]</l>, then return null.
```
mkruisselbrink, line 1531: that `read-only mode` link is wrong, no `[=`, and the `</l>` should be `</a>`.

And in general, please make sure there are no (new) bikeshed errors when building. (I really should fix the travis setup to fail the build if bikeshed fails)

index.bs
```diff
@@ -1507,6 +1505,81 @@ these steps:
...
+
+1. If the <a spec=html>[=the drag data item kind=]</a> is not <em>File</em>, then return
```
mkruisselbrink, line 1533: you don't need `<a>` and `[=`. Just using `<a>` is the same as using [= already.

index.bs
```diff
@@ -1507,6 +1505,81 @@ these steps:
...
+
+1. else if |entry| is a [=directory entry=]:
+
+  1. Let |handle| be a {{FileSystemDirectoryHandle}} associated with entry
+
+1. Return |entry|
+
+</div>
+
+Handling drag and drop of files and directories:
```
mkruisselbrink, line 1554: sorry, this should still be inside the `<div>`, just not inside the `<xmp>`.

index.bs
```diff
@@ -1507,6 +1505,81 @@ these steps:
...
+      } else if (entry.isFile) {
+        handleFileEntry(entry);
+      }
+    }
+  }
+});
+</xmp>
+</div>
+
+Issue: This currently does not block access to [=too sensitive or dangerous=] directories, to be consistent with other APIs that give access to dropped files and directories. This is inconsistent with the [=nfs handle factories=] though, so we might want to reconsider this.
```
mkruisselbrink, line 1581: sorry, i used "nfs" as shorthand in my previous comment, but the actual linkable term would be [=native file system handle factories=] I think.

index.bs
```diff
@@ -1507,6 +1506,81 @@ these steps:
...
+    FileSystemHandle? getAsFileSystemHandle();
+};
+</xmp>
+
+During a <em>drag-and-drop operation</em>, dragged file and
+directory items are associated with [=file entries=] and [=directory entries=] 
+respectively.
+
+<div class="note dominto">
+  : |handle| = item . {{getAsFileSystemEntry()}}
```
mkruisselbrink, line 1523: here too getAsFileSystemHandle

index.bs
```diff
@@ -1507,6 +1506,81 @@ these steps:
...
+  :: Returns a {{FileSystemFileHandle}} object if the dragged item is file and a {{FileSystemDirectoryHandle}} object if the dragged item is a directory.
+</div>
+
+<div algorithm>
+
+The <dfn method for=DataTransferItem>getAsFileSystemEntry()</dfn> method steps are:
```
mkruisselbrink, line 1529: here too getAsFileSystemHandle

index.bs
```diff
@@ -1507,6 +1506,81 @@ these steps:
...
+});
+elem.addEventListener('drop', e => {
+  // Prevent navigation.
+  e.preventDefault();
+
+  // Process all of the items.
+  for (const item of e.dataTransfer.items) {
+    // kind will be 'file' for file/directory entries.
+    if (item.kind === 'file') {
+      const entry = item.getAsFileSystemEntry();
```
mkruisselbrink, line 1570: here too getAsFileSystemHandle

index.bs
```diff
@@ -1507,6 +1506,81 @@ these steps:
...
+      if (entry.isDirectory) {
```
mkruisselbrink, line 1571: in the mean time isDirectory/isFile was changed to entry.kind returning 'file' or 'directory'

index.bs
```diff
@@ -1507,6 +1506,81 @@ these steps:
...
+partial interface DataTransferItem {
+    FileSystemHandle? getAsFileSystemHandle();
+};
+</xmp>
+
+During a <em>drag-and-drop operation</em>, dragged file and
+directory items are associated with [=file entries=] and [=directory entries=] 
+respectively.
+
+<div class="note dominto">
```
mkruisselbrink, line 1522: domintro instead of dominto

index.bs
```diff
@@ -1507,6 +1508,83 @@ these steps:
...
+
+1. If the {{DataTransferItem}} object is not in the <a spec=html>read/write
+    mode</a> or the <a spec=html>read-only mode</a>, then return null.
+
+1. If the <a spec=html>the drag data item kind</a> is not <em>File</em>, then return
+    null.
+
+1. Let |entry| be the [=entry=] representing the dragged file or directory.
+
+1. Set |entry|'s [=query permission steps=] to the [=native file system query permission steps=]
```
mkruisselbrink, line 1541: nit: end sentences with a .

index.bs
```diff
@@ -1507,6 +1508,83 @@ these steps:
...
+
+1. Set |entry|'s [=request permission steps=] to the [=native file system request permission steps=]
```
mkruisselbrink, line 1543: nit: end sentences with a .

index.bs
```diff
@@ -1507,6 +1508,83 @@ these steps:
...
+
+1. If |entry| is a [=file entry=]:
+
+  1. Let |handle| be a {{FileSystemFileHandle}} associated with entry
```
mkruisselbrink, line 1547: nit: end sentences with a .

index.bs
```diff
@@ -1507,6 +1508,83 @@ these steps:
...
+
+1. else if |entry| is a [=directory entry=]:
+
+  1. Let |handle| be a {{FileSystemDirectoryHandle}} associated with entry
```
mkruisselbrink, line 1551: nit: end sentences with a .

index.bs
```diff
@@ -1507,6 +1508,83 @@ these steps:
...
+
+1. Return |entry|
```
mkruisselbrink, line 1553: nit: end sentences with a .

index.bs
```diff
@@ -1507,6 +1508,83 @@ these steps:
 
 </div>
 
+## Drag and Drop ## {#drag-and-drop}
+
+<xmp class=idl>
+partial interface DataTransferItem {
+    FileSystemHandle? getAsFileSystemHandle();
```
mkruisselbrink, line 1515: thinking more about it, and considering the implementation difficulties we're running into to make it possible for this method to be synchronous, we should probably change this to return a `Promise<FileSystemHandle?>`. That way implementations have more freedom to figure out when to do certain checks. (you'll also need to update the domintro and example to work with returned promises).

mkruisselbrink (commented):
generally looks good now, but I'm thinking we should probably make this method return a promise, regardless of what we actually end up implementing.

mkruisselbrink:
> generally looks good now, but I'm thinking we should probably make this method return a promise, regardless of what we actually end up implementing.

I'd also be happy to make those changes myself, if you're too busy with the implementation side of things; just let me know.

index.bs
mkruisselbrink: Although with #200 merged, these permission bits aren't needed like this anymore. Sorry about the merge conflicts.

mkruisselbrink:
@HazimMohamed I don't think you actually uploaded your latest changes?

index.bs
```diff
@@ -1466,9 +1466,86 @@ these steps:
   1. [=/Resolve=] |p| with |result|.
 
 1. Return |p|.
+</div>
```
mkruisselbrink, line 1469: you need a blank line before the </div> for bikeshed to properly realize the markdown formatted ordered list ended.
(i.e. to silence the "FATAL ERROR: Line 1469 isn't indented enough (needs 1 indent) to be valid Markdown:" error)

index.bs
```diff
@@ -1466,9 +1466,86 @@ these steps:
...
+
+<div algorithm>
+
+The <dfn method for=DataTransferItem>getAsFileSystemHandle()</dfn> method steps are:
+
+1. Let |p| be [=a new promise=].
+
+1. Run the following steps [=in parallel=]:
 
+  1. If the {{DataTransferItem}} object is not in the <a spec=html>read/write
```
mkruisselbrink, line 1496: I'm really sorry, I know we talked about this before, but thinking about this more, checking things like the read/write mode of the DataTransferItem has to be done before we run things in parallel.

So probably easiest to do it before you even create |p|, as first two steps of this overall algorithm ```1. If ...., return [=a promise resolved with=] `null`.```

mkruisselbrink (approved):

jimmywarting:
I tried making a automated unit test. But this `handle` return undefined in chrome... Bug? Think it should have either thrown a error or returned a FileSystemFileHandle
Only works when you are actually drag n' dropping a file - how about stuff that had been created from the memory?

```js
var dt = new ClipboardEvent('').clipboardData || new DataTransfer()
dt.items.add(new File(['content'], 'sample1.txt'))
var handle = await dt.items[0].getAsFileSystemHandle() // undefined in chrome canary 
```

mkruisselbrink:
There currently is no way to have a FileSystemHandle object that somehow is related to a script-created File object (not even sure what the semantics of such a thing would be). So as such I think this is working as intended.

We should describe in the spec how this should behave though, as the current language indeed doesn't address this properly (I suspect webkitGetAsEntry might have a similar defect, both in implementation and spec?)

jimmywarting:
> So as such I think this is working as intended.

Think it should have at least thrown a error message. Undefined is not a very helpful response as to why it didn't work.

PR #195, opened 2020-07-02
Fix SVG example

RReverser:
This example was invalid, since `accept` requires a sequence as a value.

***
<a href="https://pr-preview.s3.amazonaws.com/RReverser/native-file-system/pull/195.html" title="Last updated on Jul 2, 2020, 1:49 PM UTC (9d26090)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/195/b1b9ce2...RReverser:9d26090.html" title="Last updated on Jul 2, 2020, 1:49 PM UTC (9d26090)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1402,7 +1402,7 @@ const options = {
   <l>{{FilePickerOptions/types}}</l>: [
     {
       <l>{{FilePickerAcceptType/accept}}</l>: {
-        'image/svg+xml': ['svg'],
+        'image/svg+xml': ['svg']
       }
     },
   ],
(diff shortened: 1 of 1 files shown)


RReverser:
I've sent a request to join the CG.

mkruisselbrink:
Good catch. The other option would of course be to also accept single strings as extensions; that would more closely match what the current file handling API explainer has in its examples (https://github.com/WICG/file-handling/blob/master/explainer.md). It does seem like we would want both to accept the same (although it's of course always easier to add acceptance of more than it is to reduce what we accept). @mgiuca do you have any opinions here?

mkruisselbrink:
I'll go ahead and merge this for now, I filed #197 to track possibly changing this to accept single extensions as well.

RReverser:
Sounds good, thanks.

PR #199, opened 2020-07-09
Follow changes in the storage spec.

mkruisselbrink:
This fixes various linking errors by updating to use newer terminology
from the storage spec. Also updates the text to be more consistent about
using "origin private" instead of "sandboxed".

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/199.html" title="Last updated on Jul 14, 2020, 5:21 PM UTC (11320c2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/199/fa242dd...11320c2.html" title="Last updated on Jul 14, 2020, 5:21 PM UTC (11320c2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1512,6 +1512,9 @@ The <dfn>origin private file system</dfn> is a [=storage endpoint=] whose
 <a spec=storage for="storage endpoint">types</a> are `« "local" »`,
 and <a spec=storage for="storage endpoint">quota</a> is null.
 
+Issue: Storage endpoints should be defined in [[storage]] itself, rather
+than being defined here. So merge this into the table there.
+
 Note: While user agents will typically implement this by persisting the contents of this
 [=origin private file system=] to disk, it is not intended that the contents are easily
 user accessible. Similarly there is no expectation that files or directories with names
--- a/index.bs
+++ b/index.bs
@@ -1527,7 +1527,7 @@ partial interface mixin WindowOrWorkerGlobalScope {
 };
 </xmp>
 
-Advisement: In Chrome this functionality is currently exposed as `FileSystemDirectoryHandle.getSystemDirectory({type: "sandbox"})`.
+Advisement: In Chrome this functionality was previously exposed as `FileSystemDirectoryHandle.getSystemDirectory({type: "sandbox"})`.
 This new method is available as of Chrome 85.
 
 <div class="note domintro">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1503,32 +1505,20 @@ these steps:
 
 </div>
 
-# Accessing the Sandboxed File System # {#sandboxed-filesystem}
+# Accessing the Origin Private File System # {#sandboxed-filesystem}
 
-## Sandboxed File System Concepts ## {#sandboxed-filesystem-concepts}
+The <dfn>origin private file system</dfn> is a [=storage endpoint=] whose
+<a spec=storage for="storage endpoint">identifier</a> is `"fileSystem"`,
```
inexorabletash, line 1511: Can this just use the `[=storage  endpoint/identifier=]` syntax? Is the `spec=storage` necessary?

index.bs
```diff
@@ -1538,10 +1528,11 @@ partial interface mixin WindowOrWorkerGlobalScope {
 </xmp>
 
 Advisement: In Chrome this functionality is currently exposed as `FileSystemDirectoryHandle.getSystemDirectory({type: "sandbox"})`.
+This new method is available as of Chrome 85.
```
inexorabletash, line 1531: I assume this means that the new name/semantics is in 85+ ? Maybe reword "is currently exposed" to "was exposed as .... in Chrome 84 and earlier." ?

index.bs
```diff
@@ -1550,16 +1541,26 @@ invoked, must run these steps:
...
 
-1. If |environment|'s [=environment settings object/origin=] is an [=opaque origin=],
+1. Let |map| be the result of running [=obtain a local storage bottle map=]
+   with |environment| and `"fileSystem"`. If this returns failure,
    return [=a promise rejected with=] a {{SecurityError}}.
 
-1. Let |storage| be |environment|'s [=environment settings object/origin=]'s [=storage unit=].
-1. Let |bucket| be |storage|'s [=bucket=].
+1. If |map|["root"] does not [=map/exist=]:
+  1. Set |map|["root"] to a new [=directory entry=].
```
inexorabletash, line 1549: I think this would be slightly more readable with a temp var, e.g. 

1. Let _dir_ be a new directory entry.
1. Set _dir_'s foo to bar...
1. Set _dir_'s foo to bar...
1. Set |map|["root"] to _dir_

index.bs
```diff
@@ -22,6 +22,8 @@ spec:webidl; type:dfn; text:resolve
 <pre class=anchors>
 urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262
   type: dfn; text: realm; url: realm
+urlPrefix: https://storage.spec.whatwg.org/; spec: storage
+  type: dfn; text: storage; url: site-storage
```
inexorabletash, line 26: What's this for? I don't see a `[=/storage=]` (etc) link

index.bs
inexorabletash: I guess it would be if there's another spec defining storage endpoint at the moment? Which one?

mkruisselbrink: The spec=storage lets me link to it despite it not being exported...

index.bs
mkruisselbrink: This is a term storage used to provide, it is linked to from the privacy section `Directories where the user agent stores [=storage|website storage=]`.

index.bs
mkruisselbrink: Changed it to "was previously exposed as" without explicitly saying when the old name went away, as it actually hasn't gone away yet, and will only be unavailable starting in M86/when we ship this outside of an origin trial. That hopefully is clear enough...

index.bs
mkruisselbrink: Done

index.bs
inexorabletash: Ah, thanks. I missed that variant in my search!

index.bs
inexorabletash: Thanks! (I add things to the `anchors` block in that case, whatever you prefer)

mkruisselbrink: The downside of the anchors block is that you no longer get an error if the thing you're trying to link to doesn't exist anymore.

Issue #119, opened 2019-11-20
Support write-only handles


foolip:
The inline issue in https://wicg.github.io/native-file-system/#api-filesystemhandle-querypermission states:
> the currently described API here assumes a model where it is not possible to have a write-only handle. I.e. it is not possible to have or request write access without also having read access. There definitely are use cases for write-only handles (i.e. directory downloads), so we might have to reconsider this.

https://github.com/WICG/native-file-system/issues/29 is about one use case for this, but I couldn't find an issue that would cover write-only directory downloads, so here it is.

jromy5:
The changing of the API file code

mingyc:
Started a proposal to add write mode: https://github.com/WICG/file-system-access/issues/463

PR #200, opened 2020-07-14
Rewrite the permissions logic to integrate with the permissions API.

### LINKED ISSUE #120: When is user activation / prompting required?
User activation is mentioned in two domintro boxes now:
https://wicg.github.io/native-file-system/#api-filesystemhandle-querypermission
https://wicg.github.io/native-file-system/#api-filesystemhandle-requestpermission

There are TODOs in those sections, but the phrasing "attempting to modify the file or directory this handle represents will require user activation" makes it sound like user activation might be required for `write()`, `truncate()` or `close()`. Presumably it won't be required with every call to `write()` however, so defining this well will be important to get interoperability in practice, as any difference between implementations can easily turn into a broken site in some browser.

@mustaqahmed

mkruisselbrink:
This is partially editorial (just changing how algorithms check
permissions), but also has a couple of normative changes:

- A change from `boolean writable` to a enum, with currently "read"
  and "readwrite" options. This is to keep open the possibility for
  write-only handles as discussed in #119.
- Changes from NotAllowedError to SecurityError in a couple of places,
  to better align with how other APIs behave.
- And of course the integration with navigator.permissions.query,
  although that is unlikely to be implemented in chrome any time soon,
  as a lot of the infra for that is missing in chrome.

This fixes #120.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/200.html" title="Last updated on Jul 15, 2020, 8:45 PM UTC (315004e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/200/97ba7e6...315004e.html" title="Last updated on Jul 15, 2020, 8:45 PM UTC (315004e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -133,50 +133,93 @@ run the following steps:
 
 ## Permissions ## {#permissions}
 
-Issue: Figure out some way to integrate this better with the permissions API [[permissions]],
-for example by making the entry be a part of the PermissionDescriptor.
+The <dfn for=PermissionName enum-value>"file-system"</dfn> [=powerful feature=]'s
+permission-related algorithms and types are defined as follows:
+
+: [=permission descriptor type=]
+:: {{FileSystemPermissionDescriptor}}, defined as:
+  <xmp class=idl>
+  enum FileSystemPermissionMode {
+    "read",
+    "readwrite"
+  };
+
+  dictionary FileSystemPermissionDescriptor : PermissionDescriptor {
+    required FileSystemHandle handle;
+    FileSystemPermissionMode mode = "read";
+  };
+  </xmp>
+
+: [=permission state constraints=]
+:: <div algorithm="permission state constraints">
+  To determine [=permission state constraints=] for a {{FileSystemPermissionDescriptor}} |desc|,
+  run these steps:
+
+  1. Let |entry| be |desc|.{{FileSystemPermissionDescriptor/handle}}'s [=FileSystemHandle/entry=].
+  1. If |entry| represents an [=/entry=] in an [=origin private file system=],
+    this descriptor's [=permission state=] should always be {{PermissionState/"granted"}}.
+  1. Otherwise, if |entry|'s [=entry/parent=] is not null, this descriptor's [=permission state=] should be
+    equal to the [=permission state=] for a descriptor with the same {{FileSystemPermissionDescriptor/mode}},
+    and a {{FileSystemPermissionDescriptor/handle}} representing |entry|'s [=entry/parent=].
+  1. Otherwise, if |desc|.{{FileSystemPermissionDescriptor/mode}} is {{"readwrite"}}:
+    1. Let |read state| be the [=permission state=] for a descriptor
+      with the same {{FileSystemPermissionDescriptor/handle}},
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -133,50 +133,93 @@ run the following steps:
...
-Issue: Figure out some way to integrate this better with the permissions API [[permissions]],
-for example by making the entry be a part of the PermissionDescriptor.
+The <dfn for=PermissionName enum-value>"file-system"</dfn> [=powerful feature=]'s
+permission-related algorithms and types are defined as follows:
+
+: [=permission descriptor type=]
+:: {{FileSystemPermissionDescriptor}}, defined as:
+  <xmp class=idl>
+  enum FileSystemPermissionMode {
+    "read",
```
pwnall, line 143: IndexedDB uses "readonly" and "readwrite". 

Do you think it'd be worth aligning? You'd then add "writeonly" later.

index.bs
```diff
@@ -133,50 +133,93 @@ run the following steps:
...
+  </xmp>
+
+: [=permission state constraints=]
+:: <div algorithm="permission state constraints">
+  To determine [=permission state constraints=] for a {{FileSystemPermissionDescriptor}} |desc|,
+  run these steps:
+
+  1. Let |entry| be |desc|.{{FileSystemPermissionDescriptor/handle}}'s [=FileSystemHandle/entry=].
+  1. If |entry| represents an [=/entry=] in an [=origin private file system=],
+    this descriptor's [=permission state=] should always be {{PermissionState/"granted"}}.
```
pwnall, line 160: This block has a lot of "should" in it, which seems unusual for an algorithm.

Does "should" signal a place where implementations can deviate (RFC-style) or is it an assertion? If it's the former, WDYT about something like "To determine [=permission state constraints=] for a {{FileSystemPermissionDescriptor}} |desc|, these steps **should** be run:" followed by the usual style for writing steps?

index.bs
mkruisselbrink: I initially started out with readonly. I switched to "read" since it seemed weird if query-ing for "readonly" access returns true when you have read and write access.

Maybe the we should have just "read" and "write", and specify the "mode" as a sequence in the permission descriptor. But then we'd need to do something if somebody is just requesting write access. Always requesting "read" and "write" if the website just asks for "write" would work, but makes it harder to later add true write-only access without breaking existing websites. I guess the other option would be to reject if you ask for just "write" access but that seems weird too, especially if a website already has read access.

So given that we're not doing write-only yet, this seemed the least bad of all the options. WDYT?

index.bs
mkruisselbrink: phrasing this as an algorithm felt weird to begin with. It's not really invoked as an algorithm. But it seemed at least the clearest way of expressing the relevant constraints. Probably I should just change the shoulds to musts.

index.bs
pwnall: We discussed this off-thread. Current approach SGTM, with the understanding that we'll ask for TAG's opinion during review.

index.bs
pwnall: Good idea! Thank you!

PR #203, opened 2020-07-16
Define types for "binary data" and "modification timestamp".

### LINKED ISSUE #112: Define types for concepts "binary data" and "modification timestamp"
https://wicg.github.io/native-file-system/#file-entry-binary-data
https://wicg.github.io/native-file-system/#file-entry-modification-timestamp

These two concepts are both unreferenced right now, but when they're used in algorithms it will be more important.

https://w3c.github.io/FileAPI/ speaks of byte sequences and number of milliseconds since the Unix Epoch, so that might be the right types.

https://fetch.spec.whatwg.org/#requests is a good example of how to define types for concepts by putting them in parenthesis.

mkruisselbrink:
Fixes #112

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/203.html" title="Last updated on Jul 16, 2020, 8:39 PM UTC (e80b51c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/203/e34c187...e80b51c.html" title="Last updated on Jul 16, 2020, 8:39 PM UTC (e80b51c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -79,8 +79,9 @@ other operating systems. Additionally underlying file systems might have further
 on what names are or aren't allowed, so a string merely being a [=valid file name=] is not
 a guarantee that creating a file or directory with that name will succeed.
 
-A <dfn lt="file|file entry">file entry</dfn> additionally consists of <dfn for="file entry">binary
-data</dfn> and a <dfn for="file entry">modification timestamp</dfn>.
+A <dfn lt="file|file entry">file entry</dfn> additionally consists of
+<dfn for="file entry">binary data</dfn> (a [=byte sequence=]) and a
+<dfn for="file entry">modification timestamp</dfn> (a number representing the number of milliseconds since the <a spec=FileAPI>Unix Epoch</a>).
 
 A <dfn lt="directory|directory entry">directory entry</dfn> additionally consists of a [=/set=] of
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -79,8 +79,9 @@ other operating systems. Additionally underlying file systems might have further
 on what names are or aren't allowed, so a string merely being a [=valid file name=] is not
 a guarantee that creating a file or directory with that name will succeed.
 
-A <dfn lt="file|file entry">file entry</dfn> additionally consists of <dfn for="file entry">binary
-data</dfn> and a <dfn for="file entry">modification timestamp</dfn>.
+A <dfn lt="file|file entry">file entry</dfn> additionally consists of
+<dfn for="file entry">binary data</dfn> (a [=byte sequence=]) and a
+<dfn for="file entry">modification timestamp</dfn> (a number representing the number of milliseconds since the <a spec=FileAPI>Unix Epoch</a>).
```
pwnall, line 84: (fine to address in the future)

Would it be worth a non-normative note that the format (number of milliseconds since epoch) interops with JavaScript's Date type?

mkruisselbrink: The FileAPI definition of Unix Epoch this links to kind of hints at that already... Also since this is only exposed via FileAPI currently, it is probably enough to rely on whatever text that spec has (and update it their if needed).

PR #207, opened 2020-07-16
Allow a single extension without having to wrap it in an array.

### LINKED ISSUE #197: Change type of accepted extensions from sequence to sequence or string?
Currently https://wicg.github.io/native-file-system/#dictdef-filepickeraccepttype defines accept as `record<USVString, sequence<USVString>> accept;`. It might be nicer if that were changed to `record<USVString, sequence<USVString> or USVString> accept;` to not require wrapping a single extension in an array.

This would also be more consistent with examples in the file handling API explainer (https://github.com/WICG/file-handling/blob/master/explainer.md).

mkruisselbrink:
Fixes #197

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/207.html" title="Last updated on Jul 21, 2020, 8:23 PM UTC (f5f8776)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/207/e34c187...f5f8776.html" title="Last updated on Jul 21, 2020, 8:23 PM UTC (f5f8776)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1089,7 +1089,7 @@ steps:
 <xmp class=idl>
 dictionary FilePickerAcceptType {
     USVString description;
-    record<USVString, sequence<USVString> or USVString> accept;
+    record<USVString, (USVString or sequence<USVString>)> accept;
 };
 
 dictionary FilePickerOptions {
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
I wonder if there is a better way to process the "list or single item" in the algorithm text...

PR #219, opened 2020-08-18
Require a "." to be included in the accepted extensions.

### LINKED ISSUE #213: Specifying extensions without leading "." doesn't match the File Handling API
This was brought up in https://crbug.com/829689#c84. I believe @mkruisselbrink brought this up on the File Handling API as well, at https://github.com/WICG/file-handling/issues/43.

I'm filing an issue here so we don't forget to resolve this in a way we're happy with before shipping.

mkruisselbrink:
This is more in line with how accepts in input type=file works, as well
as how the file handling API currently is specified.

Fixes #213

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/pull/219.html" title="Last updated on Aug 18, 2020, 7:45 PM UTC (8e0e0b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/native-file-system/219/ebf0210...8e0e0b9.html" title="Last updated on Aug 18, 2020, 7:45 PM UTC (8e0e0b9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1345,14 +1345,14 @@ const options = {
     {
       <l>{{FilePickerAcceptType/description}}</l>: 'Text Files',
       <l>{{FilePickerAcceptType/accept}}</l>: {
-        'text/plain': ['txt', 'text'],
-        'text/html': ['html', 'htm']
+        'text/plain': ['.txt', '.text'],
+        'text/html': ['.html', '.htm']
       }
     },
     {
       <l>{{FilePickerAcceptType/description}}</l>: 'Images',
       <l>{{FilePickerAcceptType/accept}}</l>: {
-        'image/*': ['png', 'gif', 'jpeg', 'jpg']
+        'image/*': ['.png', '.gif', '.jpeg', '.jpg']
       }
     }
   ],
@@ -1367,7 +1367,7 @@ const options = {
   <l>{{FilePickerOptions/types}}</l>: [
     {
       <l>{{FilePickerAcceptType/accept}}</l>: {
-        'image/svg+xml': 'svg'
+        'image/svg+xml': '.svg'
       }
     },
   ],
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -1384,22 +1384,25 @@ run these steps:
 1. Let |accepts options| be a empty [=/list=] of [=pairs=].
 1. [=list/For each=] |type| of |options|.{{FilePickerOptions/types}}:
   1. Let |description| be |type|.{{FilePickerAcceptType/description}}.
-  1. [=map/For each=] |type string| → |extensions| of |type|.{{FilePickerAcceptType/accept}}:
+  1. [=map/For each=] |type string| → |suffixes| of |type|.{{FilePickerAcceptType/accept}}:
     1. Let |parsedType| be the result of [=parse a MIME type=] with |typeString|.
```
inexorabletash, line 1388: Not new in this PR, but there's inconsistency between _type string_ and _typeString_ in this algorithm. The for each uses _type string_, the steps use _typeString_.

index.bs
```diff
@@ -1384,22 +1384,25 @@ run these steps:
...
     1. If |parsedType|'s [=MIME type/parameters=] are not empty, throw a {{TypeError}}.
+    1. If |suffixes| is a string:
+      1. If |suffixes| does not start with ".", throw a {{TypeError}}.
+    1. Otherwise, [=list/for each=] |suffix| of |suffixes|:
+      1. If |suffix| does not start with ".", throw a {{TypeError}}.
 
   1. Let |filter| be the following steps, given a |filename| (a [=string=]), and a |type| (a [=MIME type=]):
-    1. [=map/For each=] |type string| → |extensions| of |type|.{{FilePickerAcceptType/accept}}:
+    1. [=map/For each=] |type string| → |suffixes| of |type|.{{FilePickerAcceptType/accept}}:
     1. Let |parsedType| be the result of [=parse a MIME type=] with |typeString|.
```
inexorabletash, line 1398: Same _type string_ vs. _typeString_ inconsistency here.

inexorabletash (approved):
Normative changes LGTM, with the issue noted. Can fix here or a follow up.

Non-normative text still refers to "extensions" and should be updated carefully in a subsequent PR, e.g. saying something like _suffixes (commonly called "extensions")_ somewhere. The domintro section should also call out that the leading `.` is required.

index.bs
mkruisselbrink: hah, by making the same mistake twice I fooled bikeshed's unused variable warning...

mkruisselbrink:
> Normative changes LGTM, with the issue noted. Can fix here or a follow up.
> 
> Non-normative text still refers to "extensions" and should be updated carefully in a subsequent PR, e.g. saying something like _suffixes (commonly called "extensions")_ somewhere. The domintro section should also call out that the leading `.` is required.

I did add a mention of "." being required. Not sure about rephrasing the non-normative text to say suffixes. HTML just calls these things extensions in its description of `input type="file" accept`. But will consider what to do for a future follow-up.

PR #231, opened 2020-09-26
Fixup the example

RReverser:
In #196 `getFile` and `getDirectory` were renamed to `get*Handle` variants, but the example got skipped and still used old methods. While at it, also updated it to use the new `showOpenFilePicker` method instead of the obsolete one.

***
<a href="https://pr-preview.s3.amazonaws.com/RReverser/native-file-system/pull/231.html" title="Last updated on Jan 21, 2021, 11:45 PM UTC (089fed4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/231/58d2fe8...RReverser:089fed4.html" title="Last updated on Jan 21, 2021, 11:45 PM UTC (089fed4)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/RReverser/native-file-system/pull/231.html" title="Last updated on Nov 14, 2022, 7:28 PM UTC (2afe4c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/231/70f0a00...RReverser:2afe4c0.html" title="Last updated on Nov 14, 2022, 7:28 PM UTC (2afe4c0)">Diff</a>


index.bs
```diff
@@ -811,7 +811,7 @@ const dir_ref = current_project_dir;
 if (!dir_ref) return;
 
 // Now get a file reference by showing a file picker:
-const file_ref = await self.chooseFileSystemEntries({type: 'openFile'});
+const file_ref = await showOpenFilePicker();
```
pwnall, line 814: Was removing `self.` intentional? I think it was there to emphasize the call's receiver.

pwnall (dismissed):
Thank you very much for the fixes!

index.bs
RReverser: Yeah, I'm not sure it's useful for the example purposes. Normally global APIs are used directly without `self.` or `this.` or `window.`, it's odd for this one to stand out.

pwnall: I think this is a better form for example code. It's more explicit, and it's more likely to work when copy-pasted into an arbitrary app.

Please add the `self.` back.

RReverser: > it's more likely to work when copy-pasted into an arbitrary app

How so? I find it more confusing for beginners, because one normally doesn't write `self.fetch` or `self.setTimeout` etc. - those are used as regular global APIs, and this one doesn't seem any different.

If you're referring to shadowing of variables, then things like `var self = this;` are very common in real-world code (pre-ES6), and would break the `self.showOpenFilePicker` example above, but wouldn't break direct `showOpenFilePicker` call.

Anoesj: I've never in my life used `self`

youk: > I've never in my life used self

I've never in my life killed a rabbit. But I am pretty sure others do. I know because I've seen rabbit meat.

mkruisselbrink:
(closing and re-opening to trigger IPR hooks)

a-sully:
Apologies, I saw that the PR updated a use of `chooseFileSystemEntries` (which has since been fixed) and assumed it was irrelevant. I now see that the `getDirectory` and `getFile` updates are still relevant

a-sully (approved):

a-sully:
Thanks for the fix!

PR #251, opened 2020-11-16
Fix ignored wildcards in spec

Anoesj:
Seems like wildcard characters are being ignored in the spec.

I'm not very familiar with writing specifications, so not sure if this fixes the problem. If it doesn't, just consider this a regular "bug report".

***
<a href="https://pr-preview.s3.amazonaws.com/Anoesj/file-system-access/pull/251.html" title="Last updated on Nov 16, 2020, 6:25 PM UTC (1074f50)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/251/7b3f9f9...Anoesj:1074f50.html" title="Last updated on Nov 16, 2020, 6:25 PM UTC (1074f50)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1327,8 +1327,8 @@ and a number of MIME types and extensions (specified as a mapping of
 MIME type to a list of extensions). If no description is provided one will be generated.
 Extensions have to be strings that start with a ".".
 
-In addition to complete MIME types, "*" can be used as the subtype of a MIME type to match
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
Marked as non substantive for IPR from ash-nazg.

mkruisselbrink (approved):
Thanks for catching and fixing this!

PR #252, opened 2020-11-30
Limit what code points are allowed in suffixes.

mkruisselbrink:
There are a number of issues with what we allow in extensions that a website can provide when showing a file picker. Since the file picker (on most platforms) appends these extensions to the filename the user enters, this can result in filenames with characters we don’t want to allow/that are otherwise problematic. In particular we don't want to allow control characters or whitespace in suffixes, or filenames that end in a '.'. As such this PR adds restrictions on what characters are allowed in accepts file extensions/suffixes, as well as limiting their length to 16.

Limiting extensions to only contain alphanumeric characters, + or . still allows all
extensions in the [shared-mime-info](https://freedesktop.org/wiki/Software/shared-mime-info/) database as well as nearly all extensions in [Wikipedia's List of filename extensions](https://en.wikipedia.org/wiki/List_of_filename_extensions).

On Windows file suffixes such as `.lnk` and `.local` are also particularly dangerous to be written to. We opted not to reject these when specified as "accepted" file types in `showSaveFilePicker` though, instead leaving it up to the user agent to ignore these file suffixes, and/or block writing to these file types completely.

### Considered alternatives

- Have a bigger allow list of characters. Other characters, such as +, -, $, # and ! are probably safe in file extensions, and have been used occasionally by applications in the past. Because of how rarely these are used, we're limiting extensions to just those characters that are at least somewhat in common use.

- Have a block list rather than an allow list. This option was rejected since while we could try to enumerate all characters that might be problematic, the consequences of not blocking something we should have blocked are much worse than not allowing something we could have allowed. As such a minimal allow list was chosen instead.

- Also reject when suffixes ending in `.lnk` or `.local` appear in the accepted file types for a showSaveFilePicker or showOpenFilePicker call. We opted against this because different platforms will likely have different extensions they might want to block. Rejecting for these only on some platforms would result in unpredictability where a website might work fine on one platform but fail on another. Rejecting these on platforms where these suffixes don't have special meaning would be needlessly limiting. Giving user agents the option to ignore these file suffixes, or at least not auto-append them to file names in save dialogs still protects the user without these downsides.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/pull/252.html" title="Last updated on Dec 2, 2020, 9:24 PM UTC (ae858c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/252/94472cf...ae858c9.html" title="Last updated on Dec 2, 2020, 9:24 PM UTC (ae858c9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -96,6 +96,13 @@ other operating systems. Additionally underlying file systems might have further
 on what names are or aren't allowed, so a string merely being a [=valid file name=] is not
 a guarantee that creating a file or directory with that name will succeed.
 
+Issue: We should consider having further normative restrictions on file names that will
+never be allowed using this API, rather than leaving it entirely up to underlying file
+systems.
+
+A <dfn>valid suffix code point</dfn> is a [=code point=] that is [=ASCII alphanumeric=],
+or U+0021 (!), U+0023 (#), U+0024 ($), U+002B (+), U+002D (-), U+002E (.), or U+005F (_).
+
 A <dfn lt="file|file entry">file entry</dfn> additionally consists of
 <dfn for="file entry">binary data</dfn> (a [=byte sequence=]) and a
 <dfn for="file entry">modification timestamp</dfn> (a number representing the number of milliseconds since the <a spec=FileAPI>Unix Epoch</a>).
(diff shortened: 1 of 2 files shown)


pwnall (approved):
LGTM. Thank you for tightening this area of the spec!

Issue #144, opened 2020-01-18
Allow path suggestions to aid file system navigation


ghost:
There might be times where an app would like to suggest a path location when saving or opening a file eg. When saving a note it might be useful to allow the editor to suggest a `%documents%` file location. The thinking behind this is that it could show that path on the users system as default view in the file picker, allowing them to save or open a note with limited navigation. 

Possible path locations may include: `%documents%`, `%desktop%`, `%pictures%`, `%videos%`

mkruisselbrink:
Also related to #85 and #94, and somewhat related to #18. I agree that this is definitely something we will want, I don't think this is blocking for shipping an initial (v1) version of this API, but this set of issues I think is pretty high on the list of things we would want to add as enhancements.

mkruisselbrink:
https://github.com/WICG/file-system-access/pull/261 has a proposal for an API for this, feedback welcome.

Issue #94, opened 2019-09-25
Set initial directory to be opened with chooseFileSystemEntries 


EricSimons:
Similar to #85, but instead I think that `chooseFileSystemEntries` should accept an optional file/dir handle that it will be initially opened to.

Passing a file/dir handle ensures the browser knows exactly which location to open w/ permissions (i.e. no string parsing/etc). Using handles makes a lot of sense, as the only time a web app should be instructing `chooseFileSystemEntries` to open a specific directory is when it already knows that directory exists (i.e. it has already received the handle in a previous call to `chooseFileSystemEntries`).

Besides the use cases mentioned in #85, this will enable file tree UIs (like the one we have on [stackblitz.com](https://stackblitz.com)) to include "create file / folder" buttons that open the picker in the desired directory the user clicked in.

IMO this is incredibly important (and the only omission I've found in this spec) for providing realistic filesystem UX.

mathiasbynens:
+1 to this. My use case is an “installer”, which lets the user select a directory on their machine, to which the installer then writes a (large) number of files.

AFAICT there is currently no way to do this other than by…

1. letting the user select a directory first, resulting in a `directoryHandle`, and then
2. using `await directoryHandle.getFile(fileName, { create: true })` to get a `fileHandle` for each file.

This currently triggers two permission requests: one for 1, and one for 2. (Thankfully, not 1 for every single file in 2.) I’d much prefer being able to just ask for write permission to a user-specified directory upfront.

mathiasbynens:
+1 to this. My use case is an “installer”, which lets the user select a directory on their machine, to which the installer then writes a (large) number of files. The thing is, the desired location is always one of three paths, depending on the operating system (`navigator.platform`).

As a workaround, I currently note down these three locations, so that the user can copy-paste them into the directory picker manually… but most users probably won’t realize they can do that.

mkruisselbrink:
https://github.com/WICG/file-system-access/pull/261 has a proposal for an API for this, feedback welcome.

EricSimons:
@mkruisselbrink I just took a quick look and this proposal seems to cover the requested functionality quite well. I particularly like the ability to specify `startInParentOf` as well as `startIn`, as I can see cases where we'd want to start the picker in the parent directory for a given folder handle (but don't have a handle for the parent dir itself).

Great work to you & team! Happy to provide feedback on anything else if it'd be helpful. We're building around the Chrome FS API quite heavily for upcoming functionality on stackblitz :)

[...]

Issue #85, opened 2019-08-22
Add an id to chooseFileSystemEntries for tracking last directory


AshleyScirra:
Often applications will use `chooseFileSystemEntries` for two distinct purposes, such as opening a document, and exporting content in a different format. Typically the user will open a document from one folder (e.g. 'Documents'), and export to another (e.g. the desktop). Similarly they may open different types of content from different locations, or save different types of content to different locations.

To improve the user experience, it's useful if `chooseFileSystemEntries` remembers the last folder the user was in. This means if they tend to open files from the same folder (e.g. 'Documents'), the file picker tends to default to the right place.

However if `chooseFileSystemEntries` is used for distinct purposes, the browser has no good way of remembering the right folder. For example if an application provides options for 'Save document as...' and 'Export as PDF...', these will both use 'save as' dialogs, with little to distinguish them from the call to `chooseFileSystemEntries`. Therefore the browser cannot remember their state separately. The end result is if the user alternates between "save as" and "export", it defaults to the wrong directory, forcing the user to navigate back to the folder they want every time.

An easy way to solve this is just to add an arbitrary string "id" to `ChooseFileSystemEntriesOptions`, e.g.:

```js
chooseFileSystemEntries({
    "type": "saveFile",
    "id": "document"
});
// ... and somewhere else ...
chooseFileSystemEntries({
    "type": "saveFile",
    "id": "export"
});
```

Now the browser can always remember the right default location for each distinct use of `chooseFileSystemEntries`. It could also use the ID to save other state associated with the window.

AshleyScirra:
For reference this is basically equivalent to [FileOpenPicker.SettingsIdentifier](https://docs.microsoft.com/en-us/uwp/api/windows.storage.pickers.fileopenpicker.settingsidentifier) in the Windows UWP platform (which is also used on FileSavePicker and FolderPicker).

guest271314:
> To improve the user experience, it's useful if chooseFileSystemEntries remembers the last folder the user was in.

That can have unintended consequences https://github.com/WICG/native-file-system/issues/71 (that still persist). 

[...]

Issue #80, opened 2019-08-16
suggest a default filename when saving


jimmywarting:
I didn't find anything about saving a file with a suggested filename in the spec

You should be able to suggest a filename when you want the user to save something
`chooseFileSystemEntries({ type:'saveFile', name: 'documents.zip' })`

the unknown filename is pretty bad if you don't know the extension and you won't be able to open it.
hence why i think `saveFile` is pretty bad and why i would only use `openDirectory` with write permission instead so i can specify the filename myself, but i don't want to enforce a filename or have write permission to a hole folder, users may not want to allow it.

opusonline:
This is the only missing feature that even the download attribute is ahead of ;)

christianliebel:
> Agreed that this does seem useful. Note that you can specify `accepts` in combination with `type: 'saveFile'`, which does seem to result in the resulting file automatically getting the extension you specified.

The open file dialog seems to select the first format passed to `accept`. It would be cool if the website could also suggest the format preselection. This could be achieved by

1. deriving it from the suggested filename’s extension
2. adding the preferred format to the `ChooseFileSystemEntriesOptions` (the `OpenFileDialog` in .NET has a `FilterIndex` property allowing to preselect the format https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.openfiledialog?view=netframework-4.8#examples)

Manually reordering the array would not be very helpful, as this would also change the sorting of the formats in the file open dialog.

christianliebel:
Also, it would be interesting to know which format the user has chosen. What’s the plan here?

- Looking at the `type` of the associated `File`?
- Reversely parse it from the extension of the `FileSystemHandle`’s `name`?

Maybe this could be informally added to the spec.

jimmywarting:
Not related to the spec, but I think i found a (mac?) bug... tested out the new `showSaveFilePicker` in chrome canary with types to hint about the extension.

```js
showSaveFilePicker({
  excludeAcceptAllOption: true,
  types: [{
    accept: {
      'image/jpg': [ "jpg" ]
    }
  }]
})
```
the filename it suggest was `.jpe`
tried the same thing using jpeg... the suggested filename extension become `jfif`

other extensions like `webp` and `png` works ok

mkruisselbrink:
@jimmywarting that sounds like https://bugs.chromium.org/p/chromium/issues/detail?id=1103133, which should be fixed now (the suggested extension should now be the first script specified extension).

[...]

PR #261, opened 2021-01-14
Add initial explainer for suggestedName, startIn and related options.

mkruisselbrink:
This describes a proposal for #85, #144, #94 amd #80.

### DIFF
--- a/SuggestedNameAndDir.md
+++ b/SuggestedNameAndDir.md
@@ -159,7 +159,7 @@ const file_ref3 = await self.showOpenFilePicker({
 
 ### Interaction of `suggestedName` and accepted file types
 
-It is possible for `suggestedName` to be inconssitent with the file types a
+It is possible for `suggestedName` to be inconsistent with the file types a
 website declared as being the accepted file types for a file picker. There are
 a couple of cases worth highlighting here:
 
@@ -196,7 +196,7 @@ as such it isn't immediately obvious what should happen if all are provided.
 Our proposal is for it to be an error to provide both `startIn` and
 `startInParentOf`. It should not be an error to provide both `startIn*` and
 `id`. If both are provided, `startIn*` will specify what directory the file
-picker should start out it, and the ultimately picked directory will be recorded
+picker should start out in, and the ultimately picked directory will be recorded
 as the last selected directory for the given `id`, such that future invocations
 with that `id` but no `startIn*` value will use the directory.
 
@@ -212,9 +212,9 @@ We could also not reject if both `startIn` and `startInParentOf` are provided.
 This might have benefits if browsers only implement one of them. Websites
 could still get some of the behavior they want without having to do tricky
 feature detection to figure out which feature is and isn't supported. On the
-other hand, allowing both would mean we'd have to specify a ordering between
+other hand, allowing both would mean we'd have to specify an ordering between
 them, and unless the ordering we specify happens to match the fallback behavior
-a website wants, websites would still need to feature detect. Only allowing one
+a website wants, websites would still need to feature-detect. Only allowing one
 is simpler to reason about.
 
 We could allow file handles to be passed to `startIn` as well. This could behave
@@ -237,7 +237,7 @@ to deal with otherwise simplifies the API somewhat.
 * Chrome : Positive, authoring this explainer
 * Gecko : No signals
 * WebKit : No signals
-* Web developers : Positive, frequently requested
+* Web developers : Positive, frequently requested (See #85, #144, #94 and #80.)
 
 ## References & acknowledgements
 
--- a/SuggestedNameAndDir.md
+++ b/SuggestedNameAndDir.md
@@ -147,6 +147,12 @@ const file_ref1 = await self.showSaveFilePicker({
 const file_ref2 = await self.showSaveFilePicker({
   id: 'flower'
 });
+
+// Picker should start out in the directory `file_ref1` was picked from.
+const file_ref3 = await self.showOpenFilePicker({
+  id: 'fruit'
+});
+
 ```
 
 ## Detailed design discussion
--- a/SuggestedNameAndDir.md
+++ b/SuggestedNameAndDir.md
@@ -1,0 +1,240 @@
+# Suggested file name and location
+
+## Authors:
+
+* Marijn Kruisselbrink
+
+## Participate
+
+* [Issue tracker](https://github.com/WICG/file-system-access/issues)
+
+## Introduction
+
+When initially shipping the File System Access API we shipped a bare minimum API
+surface for showing file and directory pickers. We're now proposing adding a
+couple of frequently requested (and commonly supported in other file picker
+APIs) options to let websites provide a couple more opportunities to influence
+the behavior of the file and directory pickers.
+
+In particular this explainer addresses letting websites give suggestions for the
+name and location of files and directories that are being saved or loaded.
+
+## Goals and Use Cases
+
+The overarching goal here is to provide websites the opportunity to provide the
+file or directory picker implementation suggestions for both the initial
+directory to show in the picker, and (in the case of save dialogs) the file name
+to save a file as. Specifically to help address the following use cases:
+
+### Save an existing file in a different directory
+
+1. User opens a file from their local file system, or from some cloud provider.
+2. User then picks the "Save As" option from the web application.
+3. User saves the file in a different directory, but using the name the file
+   already had (without having to re-type the name).
+
+### Save a file using a different name in the same directory
+
+1. Users opens a file from their local file system.
+2. User then picks the "Save As" option from the web application.
+3. User saves the file using a different name, in the directory the original
+   file exists (without having to re-navigate to that directory).
+
+### Prompt a user to open a image (or video, or audio file)
+
+1. User picks "Insert image" in a web application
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
@a-sully Would you also mind taking a look? Currently this doesn't actually match what you proposed while implementing, I'd be happy to be convinced that what you've implemented makes more sense than what is described here.

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+
+The possible values for `startIn` would be:
+
+- `desktop`: The user's Desktop directory, if such a thing exists.
+- `documents`: Directory in which documents created by the user would typically be stored.
+- `downloads`: Directory where downloaded files would typically be stored.
+- `home`: The user's Home directory.
+- `music`: Directory where audio files would typically be stored.
+- `pictures`: Directory where photos and other still images would typically be stored.
+- `videos`: Directory where videos would typically be stored.
```
tomayac, line 130: macOS calls this directory "Movies".

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+There are two cases where it is useful to be able to specify a starting
+directory based on an existing handle. The first of these if a website wants to
+let a user pick a sibling to an existing file or directory. For this purpose we
+propose adding a `startInParentOf` option, that can be passed to any of the file
+and directory picker methods:
+
+```javascript
+const existing_handle = /* some FileSystemHandle, could be either a file or a directory */;
+const file_ref = await self.showSaveFilePicker({
+  startInParentOf: existing_handle
```
tomayac, line 93: I assume it's out-of-scope to make this recursive. In the example below, if you have a handle on `file_1`, there would be no way to start in `root`. This assumes that the app already as access to `root` from an earlier permission prompt.

```
-root
    |
    subfolder
      |_file_1
      |_file_2
```

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+dangerous to be allowed, implementations should be free to ignore the suggestion
+(or adjust the suggested file name to not be dangerous anymore).
+
+## Proposed API
+
+### Specifying suggested file name to save as
+
+```javascript
+const file_ref = await self.showSaveFilePicker({
+  suggestedName: 'README.md',
```
tomayac, line 72: LGTM

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+The second situation is where a website wants to prompt the user to open a file
+in a directory it already has a handle to. For example if a website has a
+"project" directory open, it might make sense for a "Save" dialog to start out
+in that directory. For this purpose we propose additionally adding a `startIn`
+option:
+
+```javascript
+const existing_dir_handle = /* some FileSystemDirectoryHandle */;
+const file_ref = await self.showSaveFilePicker({
+  startIn: existing_dir_handle
```
tomayac, line 106: LGTM

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+
+### Interaction between `startIn`, `startInParentOf` and `id`
+
+All these attributes influence what directory the file picker should start with,
+as such it isn't immediately obvious what should happen if all are provided.
+
+Our proposal is for it to be an error to provide both `startIn` and
+`startInParentOf`. It should not be an error to provide both `startIn*` and
+`id`. If both are provided, `startIn*` will specify what directory the file
+picker should start out it, and the ultimately picked directory will be recorded
```
tomayac, line 193:
```suggestion
picker should start out in, and the ultimately picked directory will be recorded
```

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+one option controls which directory to start in. There don't see to be many
+downsides to allowing both though, as recording the last used directory for
+future invocations of a file picker without a `startIn` option still seems
+useful.
+
+We could also not reject if both `startIn` and `startInParentOf` are provided.
+This might have benefits if browsers only implement one of them. Websites
+could still get some of the behavior they want without having to do tricky
+feature detection to figure out which feature is and isn't supported. On the
+other hand, allowing both would mean we'd have to specify a ordering between
```
tomayac, line 209:
```suggestion
other hand, allowing both would mean we'd have to specify an ordering between
```

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+them, and unless the ordering we specify happens to match the fallback behavior
+a website wants, websites would still need to feature detect. Only allowing one
```
tomayac, line 211:
```suggestion
a website wants, websites would still need to feature-detect. Only allowing one
```

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+the proposed API. Furthermore we would then have to specify what happens if both
+a `startIn` file, and an explicit `suggestedName` is specified.
+
+Finally, rather than having both `startIn` and `startInParentOf`, we could have
+a single option that accepts both a file or a directory handle. If a file is
+passed in, it behaves as the above described `startInParentOf`, but if a
+directory is passed in it behaves as `startIn`. This means we would lose the
+ability to start a picker in the parent directory of a given directory handle,
+but that is also one of the more edge use cases, and only having a single option
+to deal with otherwise simplifies the API somewhat.
```
tomayac, line 227: Having just a flexible `startIn` sounds a little more straightforward than having both. Another alternative could be to make file handles aware of their parent folder, so you could do `startIn: fileHandle.getParentFolder()`.

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Stakeholder Feedback / Opposition
+
+* Chrome : Positive, authoring this explainer
+* Gecko : No signals
+* WebKit : No signals
+* Web developers : Positive, frequently requested
```
tomayac, line 234:
```suggestion
* Web developers : Positive, frequently requested (See #85, #144, #94 and #80.)
```

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,246 @@
...
+// Picker should start out in the directory `file_ref1` was picked from.
+const file_ref3 = await self.showOpenFilePicker({
+  id: 'fruit'
+});
+
+```
+
+## Detailed design discussion
+
+### Interaction of `suggestedName` and accepted file types
```
a-sully, line 160: LGTM.

I looked into if we should do anything differently when the `suggestedName` does not have an extension but valid accepted extensions are specified (Ex: saving an image when you don't care about the type.
`suggestedName: 'my_pic' ... accept: {'image/*': []}`), but feel pretty strongly that what's described here is the way to go. This would technically fall into the 3rd case, meaning that we're proposing it will behave as if `excludeAcceptAllOption` was set to false.

If the `suggestedName` does not have an extension, some file pickers (ex: ChromeOS does, Linux doesn't) will automatically append the appropriate extension when `excludeAcceptAllOption = true`, but won't when it's false. This sounds nice in theory (`image/jpeg` defaults to an `.jpeg` extension), but aside from the inconsistency across platforms it can result in some weird behavior (ex: `image/*` defaults to an `.tif` extension). 

Meanwhile, requiring that a `suggestedName` have an extension specified is an arbitrary restriction which would limit use cases, while automatically appending an extension is not a great idea for the reasons listed in Considered Alternatives.

SuggestedNameAndDir.md
```diff
@@ -0,0 +1,240 @@
...
+});
+
+const file_ref2 = await self.showSaveFilePicker({
+  id: 'flower'
+});
+```
+
+## Detailed design discussion
+
+### Interaction of `suggestedName` and accepted file types
```
a-sully, line 160: Should we discuss behavior in situations where dangerous extensions are suggested? Ex: `suggestedName: 'file.lnk'`

SuggestedNameAndDir.md
mkruisselbrink: I changed it to have just a single `startIn` as my preferred option. I agree that it seems more straightforward.

Adding `handle.getParent()` or something like that comes up every now and then. We so far have resisted doing so, because it is actually not always very clear what such a method would return. Especially how it would interact with permissions is non obvious (would it only return something if you happen to already have read access to it somehow? In which case it wouldn't be very useful for this use case. But otherwise we have to be very careful not to accidentally reveal too much about the structure of directories on your disk). In general things seems much easier to reason about if "get the parent" isn't an operation that is provided by the API. If you want getParent() to just return "the directory handle (if any) this file/directory was received via getFileHandle/getDirectoryHandle" the website can just keep track of that themselves.

SuggestedNameAndDir.md
mkruisselbrink: I added a security considerations section mentioning some of this.

a-sully (approved):
LGTM

mkruisselbrink:
Merging this, to have some explainer landed. Happy to discuss more in issues etc.

jimmywarting:
is it possible to select a dir/file from `navigator.storage.getDirectory()` if you provide the the sandboxed root dir as `startIn`?

mkruisselbrink:
No, if something passed into startIn is in the sandboxed FS it is ignored.

PR #266, opened 2021-01-20
Support seeking past the end of a file.

### LINKED ISSUE #262: feature-request seek past end of file
Let writers seek past the end of a file to extend it.

I'm implementing a bittorrent client, and the requirement of keeping track of chunk files adds to complexity, reduces performance, and increases needed storage while downloading.  This is a basic functionality of many fs apis, and it's sorely missing here.

mkruisselbrink:
Rather than rejecting when trying to write past the end of a file this
changes the spec to require extending a file with some number of
0x00 (NUL) bytes instead.

This enables creating sparse files (very useful for certain file formats
such as disk images). Additionally this greatly simplifies saving content
to a file when the data to be written is received out of order (such as in
the case of bittorrent downloads). Without this functionality such a
application would have to manually make sure to resize the file either
ahead of time or when needed during writing.

Note that the behavior here would have already been possible by keeping
track of the file size, and inserting appropriate truncate() calls to
grow the file whenever trying to write past the end of a file. Not
requiring these explicit truncate calls simplifies code, and reduces the
chance of bugs in websites (accidentally shrinking files when they don't
need to, for example).

This fixes #262

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/pull/266.html" title="Last updated on Jan 23, 2021, 12:06 AM UTC (882219a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/266/65e990d...882219a.html" title="Last updated on Jan 23, 2021, 12:06 AM UTC (882219a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -976,12 +976,6 @@ runs these steps:
       1. If |input| is a {{WriteParams}} and |input|.{{WriteParams/position}} is not `undefined`,
          set |writePosition| to |input|.{{WriteParams/position}}.
       1. Let |oldSize| be |stream|.[=[[buffer]]=]'s [=byte sequence/length=].
-      1. If |writePosition| is larger than |oldSize|,
-         [=/reject=] |p| with a {{InvalidStateError}} and abort.
-
-         Issue: Not clear if this should reject, and if it does, is this really the right error type?
-         Chrome's implementation is actually inconsistent about this rejecting or not.
-
       1. If |data| is a {{BufferSource}},
          let |dataBytes| be [=get a copy of the buffer source|a copy of=] |data|.
       1. Else if |data| is a {{Blob}}:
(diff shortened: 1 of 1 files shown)


a-sully (approved):
LGTM!

domenic:
> extending a file with 0 bytes instead.

I found this very confusing, but after reading the spec change I understand you mean "some number of bytes whose value is 0", not "zero new bytes" :)

mkruisselbrink:
> > extending a file with 0 bytes instead.
> 
> I found this very confusing, but after reading the spec change I understand you mean "some number of bytes whose value is 0", not "zero new bytes" :)

Yeah, that was definitely written in a pretty confusing manner. Rephrased the description to be clearer.

PR #270, opened 2021-01-22
Update security-privacy-questionnaire.md for suggested name/dirs.

mkruisselbrink:
Making the impact of the newly proposed features in SuggestedNameAndDir.md a bit more explicit.

### DIFF
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -26,6 +26,8 @@ Furthermore, the user will be able to clear storage (storage is file handles in 
 
 Websites can also store any state they like in files they get write access to via this API. Since files written to using this API are considered to be data owned by the user, not by the application, this state would not be cleared when clearing browser data. However access to this state would be removed. If a user later picks the same files or directories again to give the website access to them, the websites will regain access to whatever state they persisted.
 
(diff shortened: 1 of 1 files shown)


security-privacy-questionnaire.md
```diff
@@ -26,6 +26,8 @@ Furthermore, the user will be able to clear storage (storage is file handles in
 
 Websites can also store any state they like in files they get write access to via this API. Since files written to using this API are considered to be data owned by the user, not by the application, this state would not be cleared when clearing browser data. However access to this state would be removed. If a user later picks the same files or directories again to give the website access to them, the websites will regain access to whatever state they persisted.
 
+Additionally, user agents could also choose to persist the last directory a file was picked from using this API on a per origin (and per purpose via the `FilePickerOption.id` option) basis. This state will not be exposed to the website, it only changes the UI that is presented to the user. A website will have no way of telling if a user picked a file in a certain directory because of this state or because the user manually navigated to the directory.
```
a-sully, line 29: > user agents could also choose to persist the last directory a file was picked from

As it's currently implemented, the last-picked directory is persisted automatically. The user agent doesn't have explicit choice in that regard. The only way to opt-out of this behavior is to specify `startIn`.

a-sully (approved):
LGTM % clarification

security-privacy-questionnaire.md
mkruisselbrink: The "user agent" is the browser implementation (i.e. chrome). In chrome we have chosen to always persist this, other implementations could make different choices.

a-sully: Whoops. Thanks for the clarification. Looks good then!

PR #283, opened 2021-03-12
Add FileSystemHandle::remove() method

### LINKED ISSUE #214: Add method to enable deleting a file without having access to its parent directory
Before #55 it was possible to delete a file by just having access to a file handle. I.e. you could get a file handle from showSaveFilePicker, decide you don't want to save after all, and delete the file. Since now deletion requires access to the parent directory this is no longer possible (as pointed out in https://crbug.com/1114923), and I think we might want to bring back a "remove self" method, at least for file handles.

a-sully:
Adds the ability for a file or directory to delete itself.

Currently, it is not possible to remove a file or directory given its handle. You must obtain the handle of the parent directory and call FileSystemDirectoryHandle::RemoveEntry. 

This would enable the common use case where you obtain a file handle from showSaveFilePicker, but then decide you don't want to save after all, and delete the file.

Fixes #214

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/pull/283.html" title="Last updated on Jun 8, 2021, 3:20 PM UTC (9429f48)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/283/e74baf7...9429f48.html" title="Last updated on Jun 8, 2021, 3:20 PM UTC (9429f48)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -340,24 +340,26 @@ these steps:
   1. If |permissionStatus| is not {{PermissionState/"granted"}},
      [=/reject=] |result| with a {{NotAllowedError}} and abort.
 
-  1. Attempt to remove |handle| from the underlying file system.
-  1. If |handle| is a [=directory entry=]:
-    1. If |handle|'s path does not exists, [=/reject=] |result| with a
+  1. Attempt to remove |entry| from the underlying file system.
+  1. If |entry| is a [=directory entry=]:
+    1. If |entry|'s path does not exists, [=/reject=] |result| with a
         {{NotFoundError}} and abort
-    1. If |handle|'s path is not a directory, [=/reject=] |result| with a
+    1. If |entry|'s path is not a directory, [=/reject=] |result| with a
         {{NotADirectoryError}} and abort
-    1. If |handle| is not an empty directory, [=/reject=] |result| with a
+    1. If |entry| is not an empty directory, [=/reject=] |result| with a
         {{NotEmptyError}} and abort
 
-  1. If |handle| is a [=file entry=]:
-    1. If |handle|'s path does not exists, [=/reject=] |result| with a
+  1. If |entry| is a [=file entry=]:
+    1. If |entry|'s path does not exists, [=/reject=] |result| with a
         {{NotFoundError}} and abort
-    1. If |handle|'s path is not a file, [=/reject=] |result| with a
+    1. If |entry|'s path is not a file, [=/reject=] |result| with a
         {{NotAFileError}} and abort
 
   1. [=/Resolve=] |result| with `undefined`.
 
 1. Return |result|.
+
+</div>
 
 ### The {{FileSystemHandle/isSameEntry()}} method ### {#api-filesystemhandle-issameentry}
 
--- a/index.bs
+++ b/index.bs
@@ -262,6 +262,8 @@ interface FileSystemHandle {
   readonly attribute FileSystemHandleKind kind;
   readonly attribute USVString name;
 
+  Promise<void> remove();
+
   Promise<boolean> isSameEntry(FileSystemHandle other);
 
   Promise<PermissionState> queryPermission(optional FileSystemHandlePermissionDescriptor descriptor = {});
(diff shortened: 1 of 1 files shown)


a-sully:
@mkruisselbrink PTAL? Particularly at the algorithm portion. Also let me know if you have more details I should add to the description. Thanks!

index.bs
```diff
@@ -315,6 +317,50 @@ and return {{FileSystemHandleKind/"directory"}} otherwise.
...
+  1. Let |permissionStatus| be the result of [=requesting file system permission=]
+     given <b>[=this=]</b> and {{"readwrite"}}.
+     If that throws an exception, [=reject=] |result| with that exception and abort.
+  1. If |permissionStatus| is not {{PermissionState/"granted"}},
+     [=/reject=] |result| with a {{NotAllowedError}} and abort.
+
+  1. Attempt to remove |entry| from the underlying file system.
+  1. If |entry| is a [=directory entry=]:
+    1. If |entry|'s path does not exists, [=/reject=] |result| with a
+        {{NotFoundError}} and abort
```
a-sully, line 346: These errors aren't linked properly. Not sure the best way to describe these

index.bs
```diff
@@ -315,6 +317,50 @@ and return {{FileSystemHandleKind/"directory"}} otherwise.
...
 associated [=FileSystemHandle/entry=].
 
+### The {{FileSystemHandle/remove()}} method ### {#api-filesystemhandle-remove}
+
+<div class="note domintro">
+  : await |handle| . {{FileSystemHandle/remove()|remove}}()
+  :: Attempts to remove the entry represented by |handle| from the underlying file system.
+
+     For files or directories with multiple hardlinks or symlinks, the entry which is deleted
+     is the entry corresponding to the path which was used to obtain the handle.
```
a-sully, line 327: @pwnall This text was snagged from your comment here. Let me know if this needs more elaboration https://github.com/WICG/file-system-access/issues/214#issuecomment-724884965

index.bs
```diff
@@ -262,6 +262,8 @@ interface FileSystemHandle {
   readonly attribute FileSystemHandleKind kind;
   readonly attribute USVString name;
 
+  Promise<void> remove();
```
mkruisselbrink, line 265: I wonder if we should have a "recurse: true/false" option analogous to the one removeEntry has. For files it wouldn't matter if you set it or not, but for directories it would make the difference between deleting the directory even when not empty, or only allowing deletion of empty directories.

index.bs
mkruisselbrink: https://heycam.github.io/webidl/#idl-DOMException-error-names lists the error names that exist on the platform (and technically everywhere in this spec I wrote things like `reject with a {{NotAllowedError}}`, that should really say `reject with a {{"NotAllowedError"}} {{DOMException}}` or something like that).

So the other error types below are not error types that actually exist, so presumably isn't what our draft implementation does either?

mkruisselbrink: Also I would probably just use similar language as the current language for `removeEntry` (or ideally make both of them more precise about the errors that can be thrown under which circumstances, I'm not terribly happy with how things currently are defined there either but being consistent is probably better than having completely different language in both places. Ideally for handles in the origin private file system we should be able to exactly define the behavior in terms of concepts defined in this spec).

a-sully: Sorry, I should have been more clear here. `{{NotFoundError}}` is not an issue, but `{{NotADirectoryError}}`, `{{NotAFileError}}`, and `{{NotEmptyError}}` are. These are file errors [1], not DOM exceptions. These are converted to DOMExceptions [2], but some information is lost in translation. The first one here is particularly bad:

```
FILE_ERROR_NOT_EMPTY_ERROR => DOMExceptionCode::kInvalidModificationError
FILE_ERROR_NOT_A_FILE      => DOMExceptionCode::kTypeMismatchError
FILE_ERROR_NOT_A_DIRECTORY => DOMExceptionCode::kTypeMismatchError
```

The DOMException translations listed here are what should tentatively go in the spec, but there's a TODO [3] for supporting custom messages which are more applicable. While that would be useful it doesn't seem like it should be a blocker here?

[1] https://source.chromium.org/chromium/chromium/src/+/main:base/files/file.h;l=88
[2] https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/fileapi/file_error.cc;l=145
[3] https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/file_system_access/file_system_access_error.cc;l=62

index.bs
a-sully: You're right, it would be odd to only allow recursive removals via removeEntry. I'll prototype this.

annevk:
If this method is still desirable (I don't remember discussing it, but it looks reasonable), could you please port this PR to https://github.com/whatwg/fs?

a-sully:
This issue has been ported to the new spec: https://github.com/whatwg/fs/pull/9

PR #287, opened 2021-03-23
Add suggested name and directory options to file picker methods.

### LINKED ISSUE #80: suggest a default filename when saving

### LINKED ISSUE #85: Add an id to chooseFileSystemEntries for tracking last directory

### LINKED ISSUE #94: Set initial directory to be opened with chooseFileSystemEntries

### LINKED ISSUE #144: Allow path suggestions to aid file system navigation

mkruisselbrink:
This adds the options as described in https://github.com/WICG/file-system-access/blob/main/SuggestedNameAndDir.md.

Fixes #85, fixes #144, fixes #94 and fixes #80

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/pull/287.html" title="Last updated on Jun 4, 2021, 5:56 PM UTC (a6f5bba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/287/e74baf7...a6f5bba.html" title="Last updated on Jun 4, 2021, 5:56 PM UTC (a6f5bba)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1153,6 +1153,8 @@ enum WellKnownDirectory {
   "videos",
 };
 
+typedef (WellKnownDirectory or FileSystemHandle) StartInDirectory;
+
 dictionary FilePickerAcceptType {
     USVString description;
     record<USVString, (USVString or sequence<USVString>)> accept;
@@ -1162,7 +1164,7 @@ dictionary FilePickerOptions {
     sequence<FilePickerAcceptType> types;
     boolean excludeAcceptAllOption = false;
     USVString id;
-    (WellKnownDirectory or FileSystemHandle)? startIn;
+    StartInDirectory? startIn;
 };
 
 dictionary OpenFilePickerOptions : FilePickerOptions {
@@ -1175,7 +1177,7 @@ dictionary SaveFilePickerOptions : FilePickerOptions {
 
 dictionary DirectoryPickerOptions {
     USVString id;
-    (WellKnownDirectory or FileSystemHandle)? startIn;
+    StartInDirectory? startIn;
 };
 
 [SecureContext]
@@ -1249,9 +1251,12 @@ To verify that an |environment| <dfn>is allowed to show a file picker</dfn>, run
 The <dfn method for=Window>showOpenFilePicker(|options|)</dfn> method, when invoked, must run
 these steps:
 
-1. Let |accepts options| be the result of [=process accept types|processing accept types=] given |options|.
-
 1. Let |environment| be <b>[=this=]</b>'s [=relevant settings object=].
+
+1. Let |accepts options| be the result of [=process accept types|processing accept types=] given |options|.
+1. Let |starting directory| be the result of [=determine the directory the picker will start in|determining the directory the picker will start in=]
+   given |options|.{{FilePickerOptions/id}}, |options|.{{FilePickerOptions/startIn}} and |environment|.
+
 1. Let |global| be |environment|'s [=environment settings object/global object=].
 1. Verify that |environment| [=is allowed to show a file picker=].
 
@@ -1266,6 +1271,8 @@ these steps:
 
      The displayed prompt should let the user pick one of the |accepts options| to filter the list of displayed files.
      Exactly how this is implemented, and what this prompt looks like is left up to the user agent.
+
+     TODO Something about |starting directory|.
 
   1. Wait for the user to have made their selection.
 
@@ -1283,6 +1290,8 @@ these steps:
          or [=/reject=] |p| with an {{AbortError}} and abort.
 
     1. Add a new {{FileSystemFileHandle}} associated with |entry| to |result|.
+
+  1. [=Remember a picked directory=] given |options|.{{FilePickerOptions/id}}, |entries|[0] and |environment|.
 
   1. Perform the <a spec=html>activation notification</a> steps in |global|'s [=Window/browsing context=].
 
@@ -1318,9 +1327,12 @@ these steps:
 The <dfn method for=Window>showSaveFilePicker(|options|)</dfn> method, when invoked, must run
 these steps:
 
-1. Let |accepts options| be the result of [=process accept types|processing accept types=] given |options|.
-
 1. Let |environment| be <b>[=this=]</b>'s [=relevant settings object=].
+
+1. Let |accepts options| be the result of [=process accept types|processing accept types=] given |options|.
+1. Let |starting directory| be the result of [=determine the directory the picker will start in|determining the directory the picker will start in=]
+   given |options|.{{FilePickerOptions/id}}, |options|.{{FilePickerOptions/startIn}} and |environment|.
+
 1. Let |global| be |environment|'s [=environment settings object/global object=].
 1. Verify that |environment| [=is allowed to show a file picker=].
 
@@ -1336,9 +1348,11 @@ these steps:
      to append to a user provided file name, but this is not required. In particular user agents are free to ignore
      potentially dangerous suffixes such as those ending in `".lnk"` or `".local"`.
 
-      If {{SaveFilePickerOptions/suggestedName}} is specified, the file picker prompt will be pre-filled with the
-     {{SaveFilePickerOptions/suggestedName}} as the default name to save as. The interaction between the
+     If |options|.{{SaveFilePickerOptions/suggestedName}} is specified, the file picker prompt will be pre-filled with the
+     |options|.{{SaveFilePickerOptions/suggestedName}} as the default name to save as. The interaction between the
      {{SaveFilePickerOptions/suggestedName}} and |accepts options| is left up to the user agent.
+
+     TODO Something about |starting directory|.
 
   1. Wait for the user to have made their selection.
 
@@ -1356,6 +1370,8 @@ these steps:
   1. Set |entry|'s [=binary data=] to an empty [=byte sequence=].
 
   1. Set |result| to a new {{FileSystemFileHandle}} associated with |entry|.
+
+  1. [=Remember a picked directory=] given |options|.{{FilePickerOptions/id}}, |entry| and |environment|.
 
   1. Perform the <a spec=html>activation notification</a> steps in |global|'s [=Window/browsing context=].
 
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
I also want to re-order the sections a bit (see the most recent commits on this branch), but temporarily reverted those changes to make the diff easier to review.

a-sully (approved):
LGTM

index.bs
```diff
@@ -1152,16 +1163,21 @@ dictionary FilePickerAcceptType {
 dictionary FilePickerOptions {
     sequence<FilePickerAcceptType> types;
     boolean excludeAcceptAllOption = false;
+    USVString id;
+    StartInDirectory? startIn;
```
inexorabletash, line 1167: Why nullable?  Prose below doesn't define null handling.

index.bs
```diff
@@ -1152,16 +1163,21 @@ dictionary FilePickerAcceptType {
...
 };
 
 dictionary OpenFilePickerOptions : FilePickerOptions {
     boolean multiple = false;
 };
 
 dictionary SaveFilePickerOptions : FilePickerOptions {
+    USVString? suggestedName;
```
inexorabletash, line 1175: Why nullable? The prose below only has "if ... is specified", doesn't handle the specified-but-null case.

index.bs
```diff
@@ -1152,16 +1163,21 @@ dictionary FilePickerAcceptType {
...
 };
 
 dictionary DirectoryPickerOptions {
+    USVString id;
+    StartInDirectory? startIn;
```
inexorabletash, line 1180: Why nullable?

index.bs
```diff
@@ -1307,13 +1340,20 @@ these steps:
...
      Exactly how this is implemented, and what this prompt looks like is left up to the user agent.
      If |accepts options| are displayed in the UI, the selected option should also be used to suggest an extension
      to append to a user provided file name, but this is not required. In particular user agents are free to ignore
      potentially dangerous suffixes such as those ending in `".lnk"` or `".local"`.
 
+     When possible, this prompt should start out showing |starting directory|.
+
+     If |options|.{{SaveFilePickerOptions/suggestedName}} is specified, the file picker prompt will be pre-filled with the
+     |options|.{{SaveFilePickerOptions/suggestedName}} as the default name to save as. The interaction between the
+     {{SaveFilePickerOptions/suggestedName}} and |accepts options| is left up to the user agent.
```
inexorabletash, line 1353: "is left to the user agent" might warrant a non-normative note giving an example of what a UA might do. Not critical though.

index.bs
```diff
@@ -1468,19 +1512,186 @@ To <dfn>validate a suffix</dfn> |suffix|, run the following steps:
 
 </div>
 
+### Starting Directory ### {#api-filepickeroptions-starting-directory}
+
+<div class="note domintro">
+The {{FilePickerOptions/id}} and {{FilePickerOptions/startIn}} fields can be specified to suggest
+the directory which the file picker opens to.
```
inexorabletash, line 1519: Sentence structure here is awkward in English.

How about: "the directory in which the file picker opens."

index.bs
```diff
@@ -1225,16 +1241,20 @@ To verify that an |environment| <dfn>is allowed to show a file picker</dfn>, run
     the selected files.
 
     Additional options can be passed to {{showOpenFilePicker()}} to indicate the types of files
-    the website wants the user to select. See [[#api-filpickeroptions-types]] for details.
+    the website wants the user to select and the preferred directory which the
```
inexorabletash, line 1244: Sentence structure is awkward in English. 

How about: "...directory in which the file picker will open."

index.bs
```diff
@@ -1468,19 +1512,186 @@ To <dfn>validate a suffix</dfn> |suffix|, run the following steps:
...
+  <l>{{FilePickerOptions/startIn}}</l>: <l>{{WellKnownDirectory/"downloads"}}</l>,  // Otherwise, start here.
+};
+</pre>
+
+</div>
+
+Advisement: The {{FilePickerOptions/startIn}} and {{FilePickerOptions/id}} options were first introduced in Chrome 91.
+
+A user agent holds a <dfn>recently picked directory map</dfn>, which is a
+[=map=] of [=/origins=] to [=path id maps=].
```
inexorabletash, line 1589: For the future: "origin" should be replaced with whatever term we come up with to represent dual-keyed storage, for browsers that support that. Do we have that term yet?

index.bs
```diff
@@ -1468,19 +1512,186 @@ To <dfn>validate a suffix</dfn> |suffix|, run the following steps:
...
+
+A <dfn>path id map</dfn> is a [=map=] of [=valid path ids=] to paths.
+
+A <dfn>valid path id</dfn> is a [=string=] where each character is [=ASCII alphanumeric=] or "_" or "-".
+
+To prevent a [=path id map=] from growing without a bound, user agents should implement
```
inexorabletash, line 1595: The spec should probably specify a lower bound for number of IDs an origin can count on.

index.bs
```diff
@@ -1468,19 +1512,186 @@ To <dfn>validate a suffix</dfn> |suffix|, run the following steps:
...
+some mechanism to limit how many recently picked directories will be remembered. This
+can for example be done by evicting least recently used entries.
+
+The <dfn enum>WellKnownDirectory</dfn> enum lets a website pick one of several well-known
```
inexorabletash, line 1599: Minor aside: Should the spec have a note that such WKDs could be virtual, e.g. on OSes that don't expose a filesystem to users but still have a concept of photos storage?

index.bs
```diff
@@ -1152,16 +1163,21 @@ dictionary FilePickerAcceptType {
...
 dictionary OpenFilePickerOptions : FilePickerOptions {
     boolean multiple = false;
 };
 
 dictionary SaveFilePickerOptions : FilePickerOptions {
+    USVString? suggestedName;
 };
 
 dictionary DirectoryPickerOptions {
+    USVString id;
```
inexorabletash, line 1179: Aside: Yay for USVString; do we have tests with a lone surrogate being mapped to U+FFFD ?

index.bs
```diff
@@ -1468,19 +1512,186 @@ To <dfn>validate a suffix</dfn> |suffix|, run the following steps:
...
+
+1. If |id| given, and is not a [=valid path id=], throw a {{TypeError}}.
+1. If |id|'s [=string/length=] is more than 32, throw a {{TypeError}}.
+
+1. Let |origin| be |environment|'s [=environment settings object/origin=].
+
+1. If |startIn| is a {{FileSystemHandle}}:
+  1. Let |entry| be |startIn|'s [=FileSystemHandle/entry=].
+  1. If |entry| does not represent an [=/entry=] in an [=origin private file system=]:
+    1. If |entry| is a [=file entry=],
```
inexorabletash, line 1636: Nit: Since this is an algorithm, it would be better to rephrase this from "If... return... if..." to "If... and ... then return..." if it can be done concisely. 

Here and below.

index.bs
```diff
@@ -1468,19 +1512,186 @@ To <dfn>validate a suffix</dfn> |suffix|, run the following steps:
...
+</div>
+
+<div algorithm>
+To <dfn>remember a picked directory</dfn>, given an optional [=string=] |id|,
+an [=/entry=] |entry|, and an [=environment settings object=] |environment|,
+run the following steps:
+
+1. Let |origin| be |environment|'s [=environment settings object/origin=].
+1. If [=recently picked directory map=][|origin|] does not [=map/exist=],
+   then set [=recently picked directory map=][|origin|] to a empty [=path id map=].
```
inexorabletash, line 1667: nit: "an empty"

index.bs
```diff
@@ -1468,19 +1512,186 @@ To <dfn>validate a suffix</dfn> |suffix|, run the following steps:
...
+    1. Let |path map| be [=recently picked directory map=][|origin|].
+    1. If |path map|[""] [=map/exists=], then return |path map|[""].
+
+1. Return a default path in a user agent specific manner.
+
+</div>
+
+<div algorithm>
+To <dfn>remember a picked directory</dfn>, given an optional [=string=] |id|,
+an [=/entry=] |entry|, and an [=environment settings object=] |environment|,
```
inexorabletash, line 1662: Nit: Is there an advantage to passing the environment in, rather than the origin?

(Doesn't really matter, but if this were code we'd want to pass in the more scoped object)

index.bs
mkruisselbrink: Done

index.bs
mkruisselbrink: Done (sort of).

index.bs
mkruisselbrink: Sorry, I'm not sure I understand what you're asking for...

index.bs
mkruisselbrink: Passing in the environment rather than just what we're using as key today seemed more flexible/future proof (probably even in code). I.e. if we'd want to change from origins to storage keys or something like that, that is now completely isolated to these algorithms, rather than spread out across all its callers. So passing in the environment means that callers don't have to worry about what is used as key.

index.bs

index.bs
inexorabletash: Right now the text is:

> If entry is a file entry, return the path on the local file system corresponding to the parent directory of entry, if such a path can be determined.

I parse this as: "If A, then B, if C" 

I believe the intent is "If (A and C), then B" but it could also be interpreted as "If A, then (if B then C, otherwise ???)"

I think it would be marginally clearer as something like:

> If entry is a file entry, and a path on the local file system corresponding to the parent directory of entry can be determined, then return that path.

But overall I think this is minor.

mkruisselbrink: Ah, yes, that makes a lot of sense. Done.

Issue #89, opened 2019-08-28
Double permission prompt when picking folder to write to


AshleyScirra:
In Chrome, if you call `window.chooseFileSystemEntries({ type: "openDirectory" })` and immediately write to the directory, the user gets a double permission prompt: first to approve "view files", and then secondly to "save changes". Writing to a chosen folder is typical usage if the app saves a folder-based user project, or needs to export multiple files.

The spec seems to make the assumption that "openDirectory" corresponds to read-only permission, so makes a permission prompt about read access. Then the first attempt to write shows a second permission prompt for write access. Ideally these could be merged in to a single prompt asking the user if they are happy for both read and write access to the folder. There doesn't seem to be a way in the spec to show the intent "open this directory for write access" though. Perhaps one should be added?

mkruisselbrink:
Sort of related to https://crbug.com/998074 as well. Yes, having a "open this directory for write access" is something worth exploring.

emarc:
We're participating in the Chrome Origin Trial, and have been trying the API out.

In short, I think the API should support asking for read+write without causing two dialogs. 

Separate dialogs make sense when different permissions are required as a result of separate actions that the user invokes, especially if the actions are optional.
But for many cases it will be immediately obvious that the application will require both read and write permissions to function. You might have to build strange user flows so that it makes sense for the user that a second popup appears, and you might have to do this in multiple places.

Ultimately the permission dialog is responsible for making the user aware of which permissions they are granting. Asking for read and write with separate dialogs does not automatically make this more clear to the user, but might instead cause “dialog desensitization” – the user is more likely to dismiss the second dialog without thought.

A longer writeup is available here: https://vaadin.com/blog/accessing-files-from-the-browser-with-the-native-file-system-api (scroll down for the actual findings)
The experiment using the API can be found here: https://vaadin.com/labs/native-fs

mkruisselbrink:
Also related to #33. We still definitely want to do this, probably won't make it in the initially shipped version, so bumping this to V2 for now.

a-sully:
[This shipped in Chrome 105](https://chromestatus.com/feature/6383970247770112?context=myfeatures) 

`showDirectoryPicker({ mode: "readwrite" })`

PR #300, opened 2021-05-28
Writable directory prompts

a-sully:
Addresses #89 

https://crbug.com/1115632

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1162,6 +1162,7 @@ dictionary SaveFilePickerOptions : FilePickerOptions {
 };
 
 dictionary DirectoryPickerOptions {
+  boolean requestWritable = false;
 };
 
 [SecureContext]
@@ -1472,8 +1473,13 @@ To <dfn>validate a suffix</dfn> |suffix|, run the following steps:
 
 <div class="note domintro">
   : |handle| = await window . {{Window/showDirectoryPicker()}}
+  : |handle| = await window . {{Window/showDirectoryPicker()}}({ {{OpenDirectoryPickerOptions/requestWritable}}: false })
   :: Shows a directory picker that lets the user select a single directory, returning a handle for
     the selected directory.
+
+  : handles = await window . {{Window/showDirectoryPicker()}}({ {{OpenDirectoryPickerOptions/requestWritable}}: true })
+  :: Shows a directory picker that lets the user select a single directory, returning a handle for
+    the selected directory. The website can ask for read and write permissions on this handle in one step.
 </div>
 
 <div algorithm>
@@ -1511,8 +1517,8 @@ these steps:
      Note: This lets a website immediately perform operations on the returned handles that
      might require user activation, such as requesting more permissions.
 
-     Issue(89): Rather than requiring the website to prompt separately for a writable directory,
-     we should provide some kind of API to request a writable directory in one step.
+     If |options|.{{OpenDirectoryPickerOptions/requestWritable}} is true, the
+     website can prompt for read and write permissions in one step.
 
   1. [=/Resolve=] |p| with |result|.
 


index.bs
```diff
@@ -1162,6 +1162,7 @@ dictionary SaveFilePickerOptions : FilePickerOptions {
 };
 
 dictionary DirectoryPickerOptions {
+  boolean requestWritable = false;
```
mkruisselbrink, line 1165: per discussion offline, what about changing this to `FileSystemPermissionMode mode = "read"`?

index.bs
```diff
@@ -1511,8 +1517,8 @@ these steps:
      Note: This lets a website immediately perform operations on the returned handles that
      might require user activation, such as requesting more permissions.
 
-     Issue(89): Rather than requiring the website to prompt separately for a writable directory,
-     we should provide some kind of API to request a writable directory in one step.
+     If |options|.{{OpenDirectoryPickerOptions/requestWritable}} is true, the
```
mkruisselbrink, line 1520: This should probably be a separate step, rather than be part of the same step that runs the activation notification steps.

We could probably also make this more formal by having a step that invokes "requesting file system permission" explicitly, rather than having this more informal language (which would also let us specify what should happen if the user rejects that permission request).

index.bs
a-sully: This has been updated.

index.bs
a-sully: I've sketched this out in the most recent patch, though I assume we might also want to consider doing this for showOpenFilePicker and showSaveFilePicker?

index.bs
```diff
@@ -1162,6 +1162,7 @@ dictionary SaveFilePickerOptions : FilePickerOptions {
 };
 
 dictionary DirectoryPickerOptions {
+  FileSystemPermissionMode mode = 'read';
```
mkruisselbrink, line 1165: nit: I think string literals in WebIDL use " rather then '.

index.bs
```diff
@@ -1472,12 +1473,17 @@ To <dfn>validate a suffix</dfn> |suffix|, run the following steps:
 
 <div class="note domintro">
   : |handle| = await window . {{Window/showDirectoryPicker()}}
+  : |handle| = await window . {{Window/showDirectoryPicker()}}({ {{OpenDirectoryPickerOptions/mode}}: 'read' })
   :: Shows a directory picker that lets the user select a single directory, returning a handle for
     the selected directory.
+
+  : handles = await window . {{Window/showDirectoryPicker()}}({ {{OpenDirectoryPickerOptions/mode}}: 'readwrite' })
+  :: Shows a directory picker that lets the user select a single directory, returning a handle for
+    the selected directory. The website can ask for read and write permissions on this handle in one step.
```
mkruisselbrink, line 1482: "the website can ask" seems weird, it is more that the user agent can combine prompts or something like that...

index.bs
```diff
@@ -1508,11 +1514,11 @@ these steps:
...
   1. Perform the <a spec=html>activation notification</a> steps in |global|'s [=Window/browsing context=].
 
-     Note: This lets a website immediately perform operations on the returned handles that
-     might require user activation, such as requesting more permissions.
-
-     Issue(89): Rather than requiring the website to prompt separately for a writable directory,
-     we should provide some kind of API to request a writable directory in one step.
+  1. Let |permissionStatus| be the result of [=requesting file system permission=]
+     given |result| and |options|.{{OpenDirectoryPickerOptions/mode}}.
+     If that throws an exception, [=reject=] |p| with that exception and abort.
```
mkruisselbrink, line 1519: I'm not sure if forwarding the exception here (and rejection with NotAllowedError below) is the right thing to do, and I'm pretty sure it at least doesn't match our current implementation, where I believe (but didn't double check) we always reject with an AbortError instead?

index.bs
```diff
@@ -1508,11 +1514,11 @@ these steps:
 
   1. Perform the <a spec=html>activation notification</a> steps in |global|'s [=Window/browsing context=].
```
mkruisselbrink, line 1515: I was thinking this should probably be after the new requesting permission steps, to match the implementation. But that probably means that the permission request steps in this algorithm should probably duplicate some of the permission request algorithm, rather than relying on calling the abstraction for it, since otherwise spec-wise it should fail if there is no current user activation at this point in the algorithm.

So the logic here then probably becomes:
 - create a FileSystemPermissionDescriptor
 - call "request permission to use" with the descriptor
 - call the "boolean permission query algorithm" to get the result of the request.

a-sully: This has been updated in the latest patch. Please confirm it's what we want :)

index.bs
```diff
@@ -1681,9 +1682,14 @@ these steps:
 
 <div class="note domintro">
   : |handle| = await window . {{Window/showDirectoryPicker()}}
+  : |handle| = await window . {{Window/showDirectoryPicker()}}({ {{DirectoryPickerOptions/mode}}: 'read' })
   :: Shows a directory picker that lets the user select a single directory, returning a handle for
     the selected directory.
 
+  : handles = await window . {{Window/showDirectoryPicker()}}({ {{DirectoryPickerOptions/mode}}: 'readwrite' })
+  :: Shows a directory picker that lets the user select a single directory, returning a handle for
+    the selected directory. The user agent can combine read and write permission requests on this handle into one prompt.
```
mkruisselbrink, line 1691: nit: perhaps mentioning something about that this will only return the directory handle if the user granted write permission might be helpful

mkruisselbrink (approved):

index.bs
a-sully: Should we mention the same thing in the `read` case above?

`Shows a directory picker that lets the user select a single directory, returning a handle for the selected directory if the user grants <mode> permission`

mkruisselbrink: I guess it wouldn't hurt. There is some overlap between this language and the language in the first paragraphs of https://wicg.github.io/file-system-access/#local-file-system-permissions, but these are more clearly meant for web developers, so repeating that info doesn't seem bad.

index.bs
```diff
@@ -1681,8 +1682,13 @@ these steps:
...
 <div class="note domintro">
   : |handle| = await window . {{Window/showDirectoryPicker()}}
+  : |handle| = await window . {{Window/showDirectoryPicker()}}({ {{DirectoryPickerOptions/mode}}: 'read' })
+  :: Shows a directory picker that lets the user select a single directory, returning a handle for
+    the selected directory if the user grants read permission.
+
+  : handles = await window . {{Window/showDirectoryPicker()}}({ {{DirectoryPickerOptions/mode}}: 'readwrite' })
   :: Shows a directory picker that lets the user select a single directory, returning a handle for
-    the selected directory.
+    the selected directory. The user agent can combine read and write permission requests on this handle into one prompt.
```
caraitto, line 1691: Can we clarify that this prompt is still separate from the directory picker?

caraitto (changes requested):

index.bs
a-sully: Good point. I've added "subsequent", which I think is enough?

PR #310, opened 2021-07-02
Add AccessHandle proposal

fivedots:
This PR adds the proposal document for AccessHandles. AccessHandles are a new proposed surface that allows us to support [Storage Foundation API's](https://github.com/WICG/storage-foundation-api-explainer) use cases through the Origin Private File System.

Further details on the origins of this proposal, and alternatives considered, can be found in the following documents: [Merging Storage Foundation API and the Origin Private File System](https://docs.google.com/document/d/121OZpRk7bKSF7qU3kQLqAEUVSNxqREnE98malHYwWec), [Recommendation for Augmented OPFS](https://docs.google.com/document/d/1g7ZCqZ5NdiU7oqyCpsc2iZ7rRAY1ZXO-9VoG4LfP7fM).

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/pull/310.html" title="Last updated on Jul 27, 2021, 5:51 PM UTC (e88a265)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/310/3309115...e88a265.html" title="Last updated on Jul 27, 2021, 5:51 PM UTC (e88a265)">Diff</a>

### DIFF
--- a/AccessHandle.md
+++ b/AccessHandle.md
@@ -46,9 +46,10 @@ a [duplex stream](https://streams.spec.whatwg.org/#other-specs-duplex) and
 auxiliary methods. The readable/writable pair in the duplex stream communicates
 with the same backing file, allowing the user to read unflushed contents.
 Another new method, *createSyncAccessHandle()*, would only be exposed on Worker
-threads. This method would offer a more buffer-based surface for reading and
-writing. The creation of AccessHandle also creates a lock that prevents write
-access to the file across (and within the same) execution contexts.
+threads. This method would offer a more buffer-based surface with synchronous
+reading and writing. The creation of AccessHandle also creates a lock that
+prevents write access to the file across (and within the same) execution
+contexts.
 
 This proposal is part of our effort to integrate [Storage Foundation
 API](https://github.com/WICG/storage-foundation-api-explainer) into File System
@@ -133,6 +134,11 @@ found in the [Appendix](#appendix).
 The reason for offering a Worker-only synchronous interface, is that consuming
 asynchronous APIs from Wasm has severe performance implications (more details
 [here](https://docs.google.com/document/d/1lsQhTsfcVIeOW80dr467Auud_VCeAUv2ZOkC63oSyKo)).
+Since this overhead is most impactful on methods that are called often, we've
+only made *read()* and *write()* synchronous. This allows us to keep a simpler
+mental model (where the sync and async handle are identical, except reading and
+writing) and reduce the number of new sync methods, while avoiding the most
+important perfomance penalties.
 
 This proposal assumes that [seekable
 streams](https://github.com/whatwg/streams/issues/1128) will be available. If
@@ -221,7 +227,7 @@ interface FileSystemSyncAccessHandle {
 
   Promise<undefined> truncate([EnforceRange] unsigned long long size);
   Promise<unsigned long long> getSize();
-  Promise<undefined> flush(); 
+  Promise<undefined> flush();
   Promise<undefined> close();
 };
 
--- a/AccessHandle.md
+++ b/AccessHandle.md
@@ -1,144 +1,112 @@
-# Adding AccessHandles to files
-
-## Authors:
-
-* Emanuel Krivoy (fivedots@chromium.org)
-* Richard Stotz (rstz@chromium.org)
-
-## Participate
-
-* [Issue tracker](https://github.com/WICG/file-system-access/issues)
-
-## Table of Contents
-
-<!-- START doctoc generated TOC please keep comment here to allow auto update -->
-<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-
-- [Introduction](#introduction)
-- [Goals and Use Cases](#goals-and-use-cases)
-- [Non-goals](#non-goals)
-- [Proposed API](#proposed-api)
-  - [New data access surface](#new-data-access-surface)
-  - [Locking semantics](#locking-semantics)
-- [Open Questions](#open-questions)
-  - [Naming](#naming)
-  - [Assurances on non-awaited consistency](#assurances-on-non-awaited-consistency)
-- [Appendix](#appendix)
-  - [AccessHandle IDL](#accesshandle-idl)
-- [Stakeholder Feedback / Opposition](#stakeholder-feedback--opposition)
-- [References & acknowledgements](#references--acknowledgements)
-
-<!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Introduction
 
-This is a proposal to support
-[Storage Foundation API's](https://github.com/WICG/storage-foundation-api-explainer)
-use cases through the Origin Private File System (OPFS), to provide a unified
-interface and a simpler model to developers.
+We propose augmenting the Origin Private File System (OPFS) with a new surface
+that brings very performant access to data. This new surface differs from
+existing ones by offering in-place and exclusive write access to a file’s
+content. This change, along with the ability to consistently read unflushed
+modifications and the availability of a synchronous variant on dedicated
+workers, significantly improves performance and unblocks new use cases for the
+File System Access API.
 
-Storage Foundation’s main objective is to support performance-sensitive
-applications through a very fast and generic storage backend. It especially
-focuses on ported WebAssembly programs for which existing storage web APIs
-don’t match the expectations of native code. There are a few design choices
-that primarily contribute to Storage Foundation’s performance:
+More concretely, we would add a *createAccessHandle()* method to the
+*FileSystemFileHandle* object. It would return an *AccessHandle* that contains
+a [duplex stream](https://streams.spec.whatwg.org/#other-specs-duplex) and
+auxiliary methods. The readable/writable pair in the duplex stream communicates
+with the same backing file, allowing the user to read unflushed contents.
+Another new method, *createSyncAccessHandle()*, would only be exposed on Worker
+threads. This method would offer a more buffer-based surface for reading and
+writing. The creation of AccessHandle also creates a lock that prevents write
+access to the file across (and within the same) execution contexts.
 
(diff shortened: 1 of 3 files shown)


AccessHandle.md
```diff
@@ -0,0 +1,247 @@
...
+  - [AccessHandle IDL](#accesshandle-idl)
+- [Stakeholder Feedback / Opposition](#stakeholder-feedback--opposition)
+- [References & acknowledgements](#references--acknowledgements)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Introduction
+
+This is a proposal to support
+[Storage Foundation API's](https://github.com/WICG/storage-foundation-api-explainer)
```
rstz, line 36: Consider reducing the focus on Storage Foundation API to improve readability

AccessHandle.md
```diff
@@ -0,0 +1,247 @@
...
+applications through a very fast and generic storage backend. It especially
+focuses on ported WebAssembly programs for which existing storage web APIs
+don’t match the expectations of native code. There are a few design choices
+that primarily contribute to Storage Foundation’s performance:
+
+*   Developers get control of when data is flushed to disk, and reads can be
+    executed consistently without flushing. This allows programs to only schedule
+    time consuming flushes when they are required for persistency, and not as a
+    precondition to operate on recently written data.
+*   Its design ensures that implementations can access data from “the renderer
```
rstz, line 50: I don't see how this is relevant for the Explainer, since the multi-process architecture is Chrome-specific and the quota system might not be part of the spec at all.

AccessHandle.md
```diff
@@ -0,0 +1,247 @@
...
+    process” (i.e. without needing to rely on a component that coordinates
+    between execution contexts) to avoid costly inter-process communication
+    (IPCs). Examples of this approach include the allocation-based quota system
+    and only allowing one open file handle per file.
+*   We avoid extra memory allocations by taking a buffer when reading.
+
+In order to bring this use cases to OPFS, we propose adding a method to
+*FileSystemFileHandle* that returns an *AccessHandle*, a new interface.
+*AccessHandles* provide the right kind of access and locking semantics to achieve
```
rstz, line 59: What do you mean by "right kind of access"?

AccessHandle.md
```diff
@@ -0,0 +1,247 @@
...
+[Origin Private File System](https://wicg.github.io/file-system-access/#sandboxed-filesystem),
+and doesn't currently consider changes to the rest of File System Access API or
+how files in the host machine are accessed.
+
+## Proposed API
+
+### New data access surface
+
+```javascript
+//In all contexts
```
rstz, line 104: Consider adding whitspace after //

AccessHandle.md
```diff
@@ -0,0 +1,247 @@
...
+const handle = await file.createAccessHandle();
+await handle.writable.getWriter().write(buffer);
```
rstz, line 106: The "writable" here is a bit confusing, since it is not the writable that's created by "createWritable()", especially since these other writables are mentioned in Line 159. If you cannot rename them, maybe

AccessHandle.md
```diff
@@ -0,0 +1,247 @@
...
+It  would return an *AccessHandle* that contains a
+[duplex stream](https://streams.spec.whatwg.org/#other-specs-duplex) and
+auxilliary methods. The readable/writable pair in the duplex stream would
+communicate with the same backing file, allowing the user to read unflushed
+contents. Another new method, *createSyncAccessHandle()*, would be only exposed
+on Worker threads. This method would offer a more buffer based surface for read
+and write. An IDL description of the new interface can be found in the
+[Appendix](#appendix).
+
+The reason for offering a Worker-only synchronous interface, is that consuming
```
rstz, line 127: I think the comma is not needed?

AccessHandle.md
```diff
@@ -0,0 +1,247 @@
...
+asynchronous APIs from Wasm has severe performance implications (more details
+[here](https://docs.google.com/document/d/1lsQhTsfcVIeOW80dr467Auud_VCeAUv2ZOkC63oSyKo)).
+We've opted for slightly different interfaces between the async and sync
+versions, because there currently isn’t support for synchronous streams. The
+amount of effort that would be required to spec them in a way that makes them
+compatible with asynchronous streams would be prohibitively high, and likely
+not worth it to support a single API.  Furthermore, we hope to discourage the
+use of the sync surface unless a developer is dealing with legacy Wasm ports,
+and therefore there is not much benefit to adding them.
```
rstz, line 136: Could you rephrase that? It sounds like there is not much benefit in the sync interface at all

AccessHandle.md
```diff
@@ -0,0 +1,247 @@
...
+
+This proposal assumes that
+[seekable streams](https://github.com/whatwg/streams/issues/1128) will be
+available. If this doesn’t happen, We can emulate the seeking behavior by
```
rstz, line 140: We -> we

AccessHandle.md
```diff
@@ -0,0 +1,247 @@
...
+} catch(e) {
+  //This catch will always be executed, since there is an open access handle
+}
+await handle1.close();
+//Now a new access handle may be created
+```
+
+In order to avoid multiple contexts modifying a file at the same time, locking
+semantics would be added to the new surface. At any given time, and across
+execution contexts, there would only be either a single *AccessHandle* per
```
rstz, line 158: or SyncAccessHandle?

AccessHandle.md
fivedots: I've rewritten a big part of the doc, could you PTAL?

AccessHandle.md
fivedots: Removed.

AccessHandle.md
fivedots: Removed.

AccessHandle.md
fivedots: Done.

AccessHandle.md
fivedots: Sadly that name is part of the definition of a duplex stream, so there is benefit to keeping it. I think there's a part of yor comment missing, what is your suggestion?

AccessHandle.md
fivedots: Removed.

AccessHandle.md
fivedots: Done.

AccessHandle.md
fivedots: Done.

AccessHandle.md
fivedots: Done.

rstz (approved):
LGTM modulo comments

AccessHandle.md
```diff
@@ -0,0 +1,206 @@
...
+[Exposed=DedicatedWorker]
+interface FileSystemSyncAccessHandle {
+  unsigned long long read([AllowShared] BufferSource buffer,
+                             FilesystemReadWriteOptions options);
+  unsigned long long write([AllowShared] BufferSource buffer,
+                              FilesystemReadWriteOptions options);
+
+  Promise<undefined> truncate([EnforceRange] unsigned long long size);
+  Promise<unsigned long long> getSize();
+  Promise<undefined> flush();
```
mkruisselbrink, line 193: iirc, in the NativeIO/Storage Foundation API these were all synchronous as well. Could you elaborate on what you learned that made you change the API shape in this newer proposal?

AccessHandle.md
```diff
@@ -0,0 +1,206 @@
...
+    relying on the new surface's performance and direct buffered access to
+    offload sound segments to disk instead of holding them in memory.
+*   Provide a fast and persistent [Emscripten](https://emscripten.org/)
+    filesystem to act as generic and easily accessible storage for Wasm.
+
+## Non-goals
+
+This proposal is focused only on additions to the [Origin Private File
+System](https://wicg.github.io/file-system-access/#sandboxed-filesystem), and
+doesn't currently consider changes to the rest of File System Access API or how
```
mkruisselbrink, line 54: While I know you're focusing on the origin private use cases specifically for this proposal, I think it would be worthwhile to explore what parts of this API could be exposed on arbitrary file handles as well. I believe there are a number of use cases that could be addressed by adding a couple of options to the createAccessHandle methods to slightly tweak its behavior.

Straw-man proposal `createAccessHandle({inPlace: true, locking: true})` would match what you've described here, and be only available for origin private files. 

On the other hand inPlace: false could then work with the same API surface for arbitrary file handles. While that would of course duplicate some API surface from the existing createWritable method, being able to read what has been written before it is flushed/comitted seems useful, and if this new API will be built on top of some better seekable streams primitive that might be a benefit as well.

Not sure if a non-locking version for the origin private file system makes sense, or is desirable, but for arbitrary files I think it would be useful to be able to create both locking and non-locking access handles.

While all that can probably be added after the fact, we can't really change the defaults for such options easily, so while it might perhaps make sense to initially only support a "inPlace: true, locking: true" mode (and throw/reject for any other options), I don't think that should be the default flags if we actually want to pursue a more unified API.

index.bs
```diff
@@ -473,7 +473,7 @@ Advisement: In the Origin Trial as available in Chrome 82, createWritable replac
   :: Returns a {{FileSystemWritableFileStream}} that can be used to write to the file. Any changes made through
      |stream| won't be reflected in the file represented by |fileHandle| until the stream has been closed.
      User agents try to ensure that no partial writes happen, i.e. the file represented by
-     |fileHandle| will either contains its old contents or it will contain whatever data was written
+     |fileHandle| will either contain its old contents or it will contain whatever data was written
```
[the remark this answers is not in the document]
fivedots: Removed the extraneous commit

AccessHandle.md
fivedots: That's a good point, I agree.

I've added a "mode" parameter to createAccessHandle, and an open question describing why it's there and what other values it could have. Please take a look and let me know if this covers what you meant!

PR #317, opened 2021-08-06
Add FileSystemHandle::move() method

a-sully:
See `MoveAndRename.md`

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/pull/317.html" title="Last updated on Sep 13, 2021, 8:19 PM UTC (a7bef63)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/317/9429f48...a7bef63.html" title="Last updated on Sep 13, 2021, 8:19 PM UTC (a7bef63)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/pull/317.html" title="Last updated on Jan 13, 2022, 6:34 PM UTC (250cdba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/317/9429f48...250cdba.html" title="Last updated on Jan 13, 2022, 6:34 PM UTC (250cdba)">Diff</a>

### DIFF
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -232,12 +232,10 @@ const file_ref = await dir_ref.getFile('foo.js');
 // Get a subdirectory.
 const subdir = await dir_ref.getDirectory('bla', {create: true});
 
-// No special API to create copies, but still possible to do so by using
-// available read and write APIs.
-const new_file = await dir_ref.getFile('new_name', {create: true});
-const new_file_writer = await new_file.createWritable();
-await new_file_writer.write(await file_ref.getFile());
-await new_file_writer.close();
+// Rename a file and/or move it to another directory
+await file_ref.move('new_name');
+await file_ref.move(other_dir_ref);
+await file_ref.move(dir_ref, 'newer_name');
 
 // Or using streams:
 const copy2 = await dir_ref.getFile('new_name', {create: true});
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -233,7 +233,7 @@ const file_ref = await dir_ref.getFile('foo.js');
 const subdir = await dir_ref.getDirectory('bla', {create: true});
 
 // Rename a file and/or move it to another directory
-await file_ref.move('new_name');
+await file_ref.rename('new_name');
 await file_ref.move(other_dir_ref);
 await file_ref.move(dir_ref, 'newer_name');
 
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -233,9 +233,16 @@ const file_ref = await dir_ref.getFile('foo.js');
 const subdir = await dir_ref.getDirectory('bla', {create: true});
 
 // Rename a file and/or move it to another directory
-await file_ref.rename('new_name');
-await file_ref.move(other_dir_ref);
-await file_ref.move(dir_ref, 'newer_name');
+await file_ref.move('new_name');             // Rename the file.
+await file_ref.move(other_dir_ref);          // Move to another directory.
+await file_ref.move(dir_ref, 'newer_name');  // Move and rename.
+
+// No special API to create copies, but still possible to do so by using
+// available read and write APIs.
+const new_file = await dir_ref.getFile('new_name', {create: true});
+const new_file_writer = await new_file.createWritable();
+await new_file_writer.write(await file_ref.getFile());
+await new_file_writer.close();
 
 // Or using streams:
 const copy2 = await dir_ref.getFile('new_name', {create: true});
--- a/MoveAndRename.md
+++ b/MoveAndRename.md
@@ -1,0 +1,164 @@
+# Adding FileSystemHandle::move() and FileSystemHandle::rename() methods
+
+## Authors:
+
+* Austin Sullivan (asully@chromium.org)
+
+## Participate
+
+* [Issue tracker](https://github.com/WICG/file-system-access/issues)
+
+## Table of Contents
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Adding FileSystemHandle::move() and FileSystemHandle::rename() methods](#adding-filesystemhandlemove-and-filesystemhandlerename-methods)
+  - [Authors:](#authors)
+  - [Participate](#participate)
+  - [Table of Contents](#table-of-contents)
+- [Overview](#overview)
+- [API Surface](#api-surface)
+  - [Open Questions - Naming](#open-questions---naming)
+- [Handling Moves to Other Directories](#handling-moves-to-other-directories)
+  - [(1) Moving to a Non-Local File System](#1-moving-to-a-non-local-file-system)
+  - [(2) Moving out of the OPFS](#2-moving-out-of-the-opfs)
+  - [(3) Moving to a Directory on the Local File System](#3-moving-to-a-directory-on-the-local-file-system)
+  - [Open Questions - Move](#open-questions---move)
+    - [Should behavior be different when moving a file vs moving a directory?](#should-behavior-be-different-when-moving-a-file-vs-moving-a-directory)
+    - [What should happen if the operation cannot be completed successfully?](#what-should-happen-if-the-operation-cannot-be-completed-successfully)
+- [Implementation Notes](#implementation-notes)
+- [Alternatives Considered](#alternatives-considered)
+  - [Only support "rename" functionality](#only-support-rename-functionality)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Overview
+
+Currently, the API does not support an efficient way to move or rename files or
+directories. This requires creating a new file/directory, copying over data
+(recursively, in the case of a directory), and removing the original. This
+process is slow, error prone (e.g. disk full), and can require re-uploading
+files in some cases (e.g. Google Drive files on ChromeOS).
+
+We propose adding new `FileSystemHandle::rename()` and
+`FileSystemHandle::move()` methods. `rename()` guarantees atomic moves of a file
+or directory without the need to duplicate data. `move()` only offers atomic
+moves of a file or directory if it is moved within the same file system, while
+moves to non-local file systems, will not be guaranteed to be atomic and may
+involve duplicating data. See the [Open Questions](#open-questions---move).
+
+# API Surface
+We propose adding a `rename()` method and a `move()` method with two variations.
+
+`rename()` will rename a file or directory. It is guaranteed to be atomic.
+
+The first `move()` variation allows a user to specify a `destination_directory`
+to move to, which may or may not be on the same file system. The name of the
+original file or directory is retained. This is not guaranteed to be atomic,
+since the destination directory may be on a different file system and/or may be
+subject to other checks (ex: Safe Browsing in Chrome).
+
+The second `move()` variation allows a user to specify a `new_entry_name` as
+well as a `destination_directory` to move to. The same (lack of) atomicity
+guarantees apply here as the first `move()` variation.
+
+```
+interface FileSystemHandle {
+    ...
+
+    Promise<void> rename(USVString new_entry_name);
+    Promise<void> move(FileSystemDirectoryHandle destination_directory);
+    Promise<void> move(FileSystemDirectoryHandle destination_directory, USVString new_entry_name);
+
+    ...
+};
+```
+
+## Open Questions - Naming
+- Should we prefer `move()` to move verbose alternatives, such as `moveTo()` or
+  `moveToDirectory()`?
+- Should we bother guaranteeing (in the spec) atomicity to all handles `move()`d
+  on the same file system, or only for `rename()`d handles?
+
+# Handling Moves to Other Directories
+
+There are three cases to consider when moving to other directories:
+
+## (1) Moving to a Non-Local File System
+Moving across file systems cannot be guaranteed to be atomic if the operation
+cannot be completed successfully (ex: network connection issues, lack disk
+space, etc.). This may result in partial writes to the target directory.
+
+Note that this approach is no worse than what is available currently.
+
+## (2) Moving out of the OPFS
+
+[#310](https://github.com/WICG/file-system-access/pull/310) exposes a strong use
+case for fast, atomic moves to other directories on the same file system. Files
+can be written efficiently in the Origin Private File System using an
+AccessHandle, then efficiently moved to a directory of the user's choice.
+However, writes using an AccessHandle are not subjected to Safe Browsing checks
+in Chrome. Moving files out of the OPFS to a user's directory will require
+running Safe Browsing checks on all moved files.
+
+For example, moving a large directory across the OPFS -> file system boundary
+means locking and performing checks (Safe Browsing in Chrome) for an unbounded
+number of files.
+
+## (3) Moving to a Directory on the Local File System
+
+This is the most straightforward case. Every `rename()` falls in this case. The
+operation should be atomic and fast, since no Safe Browsing checks need to be
+performed. However, providing these guarantees for `move()` in only this case
+may be confusing.
+
+## Open Questions - Move
+
+### Should behavior be different when moving a file vs moving a directory?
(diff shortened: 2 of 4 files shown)


a-sully:
Note that at this point the changes to the explainer are still very much WIP.

pwnall:
@a-sully Could you please hash out an alternative where the API throws on non-atomic moves?

I think it's worth discussing the benefits (no performance cliff, more flexibility in handling errors for the app, less complexity for browser implementers translating to less room for implementation-specific behavior) versus the costs (potential for errors not caught in development, apps need code to handle the cross-filesystem case).

Also, we should discuss doing different things for moving files vs directories. Moving a large directory across the OPFS -> file system boundary means locking and possibly performing checks (Safe Browsing in Chrome) for an unbounded number of files.

a-sully:
@pwnall I've updated the description to talk about the things you mentioned. To be clear, are you talking about throwing on all moves which we cannot guarantee to be atomic (moves to non-local filesystems) or only when we attempt said move and it fails?

EXPLAINER.md
```diff
@@ -232,12 +232,10 @@ const file_ref = await dir_ref.getFile('foo.js');
...
-const new_file = await dir_ref.getFile('new_name', {create: true});
-const new_file_writer = await new_file.createWritable();
-await new_file_writer.write(await file_ref.getFile());
-await new_file_writer.close();
+// Rename a file and/or move it to another directory
+await file_ref.rename('new_name');
+await file_ref.move(other_dir_ref);
+await file_ref.move(dir_ref, 'newer_name');
 
 // Or using streams:
```
pwnall, line 240: I think that lines 242-244 are showing another approach for the copying task in lines 235-240. If we're deleting the copy example, I think we need to delete those lines as well.

With that being said, there's a difference between copying and moving, so I'm not 100% sure we need to remove the copying examples as we introduce the new example for moving.

pwnall (approved):
The explainer and spec changes LGTM, given the idea we're proposing.

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Overview
+
+Currently, the API does not support an efficient way to move or rename files or
+directories. This requires creating a new file/directory, copying over data
+(recursively, in the case of a directory), and removing the original. This
+process is slow, error prone (e.g. disk full), and can require re-uploading
+files in some cases (e.g. Google Drive files on ChromeOS).
```
tomayac, line 42:
```suggestion
files in some cases (e.g. Google Drive files on Chrome OS).
```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+
+We propose adding new `FileSystemHandle::rename()` and
+`FileSystemHandle::move()` methods. `rename()` guarantees atomic moves of a file
+or directory without the need to duplicate data. `move()` only offers atomic
+moves of a file or directory if it is moved within the same file system, while
+moves to non-local file systems, will not be guaranteed to be atomic and may
```
tomayac, line 48:
```suggestion
moves to non-local file systems will not be guaranteed to be atomic and may
```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+involve duplicating data. See the [Open Questions](#open-questions---move).
+
+# API Surface
```
tomayac, line 51:
```suggestion
# API Surface

```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+We propose adding a `rename()` method and a `move()` method with two variations.
+
+`rename()` will rename a file or directory. It is guaranteed to be atomic.
+
+The first `move()` variation allows a user to specify a `destination_directory`
+to move to, which may or may not be on the same file system. The name of the
+original file or directory is retained. This is not guaranteed to be atomic,
+since the destination directory may be on a different file system and/or may be
+subject to other checks (ex: Safe Browsing in Chrome).
```
tomayac, line 60:
```suggestion
subject to other checks (e.g. Safe Browsing in Chrome).
```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+
+The second `move()` variation allows a user to specify a `new_entry_name` as
+well as a `destination_directory` to move to. The same (lack of) atomicity
+guarantees apply here as the first `move()` variation.
```
tomayac, line 64:
```suggestion
guarantees apply here as with the first `move()` variation.
```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+
+    Promise<void> rename(USVString new_entry_name);
+    Promise<void> move(FileSystemDirectoryHandle destination_directory);
+    Promise<void> move(FileSystemDirectoryHandle destination_directory, USVString new_entry_name);
+
+    ...
+};
+```
+
+## Open Questions - Naming
```
tomayac, line 78:
```suggestion
## Open Questions - Naming

```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+- Should we prefer `move()` to move verbose alternatives, such as `moveTo()` or
+  `moveToDirectory()`?
+- Should we bother guaranteeing (in the spec) atomicity to all handles `move()`d
+  on the same file system, or only for `rename()`d handles?
+
+# Handling Moves to Other Directories
+
+There are three cases to consider when moving to other directories:
+
+## (1) Moving to a Non-Local File System
```
tomayac, line 88:
```suggestion
## (1) Moving to a Non-Local File System

```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+Moving across file systems cannot be guaranteed to be atomic if the operation
+cannot be completed successfully (ex: network connection issues, lack disk
```
tomayac, line 90:
```suggestion
cannot be completed successfully (e.g. network connection issues, lack disk
```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+space, etc.). This may result in partial writes to the target directory.
+
+Note that this approach is no worse than what is available currently.
+
+## (2) Moving out of the OPFS
+
+[#310](https://github.com/WICG/file-system-access/pull/310) exposes a strong use
+case for fast, atomic moves to other directories on the same file system. Files
+can be written efficiently in the Origin Private File System using an
+AccessHandle, then efficiently moved to a directory of the user's choice.
```
tomayac, line 100:
```suggestion
`AccessHandle`, then efficiently moved to a directory of the user's choice.
```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+However, writes using an AccessHandle are not subjected to Safe Browsing checks
```
tomayac, line 101:
```suggestion
However, writes using an `AccessHandle` are not subjected to Safe Browsing checks
```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+## (3) Moving to a Directory on the Local File System
+
+This is the most straightforward case. Every `rename()` falls in this case. The
+operation should be atomic and fast, since no Safe Browsing checks need to be
+performed. However, providing these guarantees for `move()` in only this case
+may be confusing.
+
+## Open Questions - Move
+
+### Should behavior be different when moving a file vs moving a directory?
```
tomayac, line 118:
```suggestion
### Should behavior be different when moving a file vs moving a directory?

```

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+  - Atomicity is only guaranteed in (2) and (3), though is likely true for (1).
+  - Speed is only guaranteed in (3).
+- For a directory:
+  - Atomicity is only easily guaranteed in (3). Atomicity for (2) would be
+    complex (likely requiring locking each file as an unbounded number of files
+    are scanned) and may not be the best option.
+  - Speed is only guaranteed in (3), but the implications of this for (1) and
+    (2) is much greater than when only moving a file.
+
+### What should happen if the operation cannot be completed successfully?
```
tomayac, line 129:
```suggestion
### What should happen if the operation cannot be completed successfully?

```

tomayac (commented):
Just a quick drive-by review, since I will need to document this on https://web.dev/file-system-access/. Do you have a Chromium bug for me to subscribe to and see when this ships (or can be tested) and thus needs documentation?

MoveAndRename.md
```diff
@@ -0,0 +1,164 @@
...
+
+## (2) Moving out of the OPFS
+
+[#310](https://github.com/WICG/file-system-access/pull/310) exposes a strong use
+case for fast, atomic moves to other directories on the same file system. Files
+can be written efficiently in the Origin Private File System using an
+AccessHandle, then efficiently moved to a directory of the user's choice.
+However, writes using an AccessHandle are not subjected to Safe Browsing checks
+in Chrome. Moving files out of the OPFS to a user's directory will require
+running Safe Browsing checks on all moved files.
```
mikewest, line 103: Will this also instruct the operating system to stamp the file with a "mark of the web" (e.g. the work behind https://source.chromium.org/chromium/chromium/src/+/main:components/services/quarantine/quarantine.h;drc=0e7c838c357344ffe4afc1b74c500df3c4caeee9;l=70 in Chromium)?

a-sully: Yes. The process for moving files should be the same as `FileSystemAccessFileWriterImpl`:

1) [scan the file](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/file_system_access/file_system_access_file_writer_impl.cc;l=375;bpv=0;bpt=1)
2) [move the file](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/file_system_access/file_system_access_file_writer_impl.cc;l=437;bpv=0;bpt=1)
3) [add"mark of the web"](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/file_system_access/file_system_access_file_writer_impl.cc;l=495;bpv=0;bpt=1)

The eventual spec language here would have to be more general, since "Safe Browsing" is a Chrome-specific concept anyways

mikewest: Ok, great. I'd suggest mentioning that explicitly... I agree that the language could be more general, but mentioning the kinds of security checks that might be necessary seems pretty reasonable to me, and might help others reading this later understand the requirements more clearly.

(Also, "Safe Browsing" isn't Chrome-specific. :) AFAIR, Safari and Firefox both use it, though to varying degrees, and with varying implementations (proxies, etc).)

mikewest (approved):

MoveAndRename.md
```diff
@@ -0,0 +1,169 @@
...
+  - This may not be possible. Ex: loss of network connection.
+  - Adds lots of complexity.
+- Abort the operation and throw an error
+  - Less complexity.
+  - No performance cliff.
+  - More flexibility in handling errors for the app.
+
+# Implementation Notes
+
+- Write permission needs to have been granted to both the source and destination
```
engedy, line 151: To double-check: This means that the `rename` operation will require write permissions on the containing directory; meaning that giving write permission to file F or directory D using the file chooser permission UX will never allow renaming F or D, right?

a-sully: Correct. Renaming a file or directory will require write permission to its parent directory.

a-sully:
> Just a quick drive-by review, since I will need to document this on https://web.dev/file-system-access/. Do you have a Chromium bug for me to subscribe to and see when this ships (or can be tested) and thus needs documentation?

Just created a launch bug: https://crbug.com/1249662

MoveAndRename.md
engedy: Sounds good! Let's spell this out in the explainer!

mkruisselbrink: Having said that, I think we could consider an extension of this where giving write permission to files F and G using the file chooser permission UX will allow renaming from F to G as well (well, not sure if write permission on the source file itself should be enough, I guess that depends on if "self deletion" is something that we want to allow or not). But it seems that write permission on the destination file as opposed to the destination folder should conceptually be enough to allow a move to that file.

engedy: It sounds reasonable to me to define the permission requirements for moving /dir1/file1 to /dir2/file2 as:
 a) Write permission on directory /dir1,
AND
 b) Write permission on directory /dir2 or write permission on file /dir2/file2, making sure to correctly apply MotW & co.

hcldan:
@a-sully 
The https://web.dev/file-system-access/
article makes it sound like you can only rename files and move directories... at least that's how I read it.  Maybe change the wording?  Unless that's how it works... in which case... why?

> Files and folders can be renamed or moved to a new location by calling rename() or move() on the FileSystemFileHandle or FileSystemDirectoryHandle respectively.

tomayac:
This is what's in the article:

> Files and folders can be renamed or moved to a new location by calling `rename()` or `move()` on the `FileSystemFileHandle` or `FileSystemDirectoryHandle` respectively. For `move()`, the first parameter is a `FileSystemDirectoryHandle`.

What do you want to change?

hcldan:
> Files and folders can be renamed or moved to a new location by calling rename() or move() on the FileSystemHandle.

MoveAndRename.md
a-sully: Updated this section to mention MOTW explicitly, and also included a new "Open Question" about how to handle MOTW in same-file-system moves

MoveAndRename.md
a-sully: Updated the doc to allow for write permission on the destination file, as described above.

a-sully:
@tomayac we actually just landed a change which consolidates the `move()` and `rename()` methods into one overloaded `move()` method. I've uploaded a new patch on this PR with the changes.

tomayac:
Please see https://github.com/GoogleChrome/web.dev/pull/7172 with these changes.

hcldan:
FWIW @a-sully I'm getting an error in canary(Version 99.0.4828.0)
image


This was the call:
image

hcldan:
image

image

Not sure I understand this one though...

hcldan:
can you not move dirs with things in them?

a-sully:
@hcldan it looks like there's been some miscommunication on our side. Due to all of the open questions regarding cross-file-system moves, `move()` is temporarily disabled for directories (hence the abort errors) and moves outside of the Origin-Private File System. This feature is currently only available in the OPFS to those with the AccessHandles origin trial enabled.

@tomayac would you mind updating web.dev to reflect this? (see the launch bug.)

We would like to support this feature more broadly eventually, but to do so we need to address the open questions mentioned in this PR. Apologies for the confusion.

hcldan:
But both of these dirs are within the OPFS.. they aren't cross filesystem... Even those dir moves are disabled?

a-sully:
Yes, directory moves within OPFS are (annoyingly) not atomic so those have been temporarily blocked, as well.

hcldan:
> 

> temporarily blocked

How temporarily?

tomayac:
> @tomayac would you mind updating web.dev to reflect this? (see the launch bug.)

Done: https://github.com/GoogleChrome/web.dev/pull/7172/commits/0f2f16177611850bd16a060728e815e044eff3e9.

hcldan:
> @hcldan it looks like there's been some miscommunication on our side. Due to all of the open questions regarding cross-file-system moves, `move()` is temporarily disabled for directories (hence the abort errors) and moves outside of the Origin-Private File System. This feature is currently only available in the OPFS to those with the AccessHandles origin trial enabled.
> 
> @tomayac would you mind updating web.dev to reflect this? (see the launch bug.)
> 
> We would like to support this feature more broadly eventually, but to do so we need to address the open questions mentioned in this PR. Apologies for the confusion.

@a-sully Thank you for this statement, but it brings up another question for me.
When chrome 99 ships there won't be a flag anymore to rope off AccessHandles.  Will this limitation ship in chrome 99, or do you expect there will be a fix for this "soon" (before chrome 99 ships)?

annevk:
@a-sully could you please rebase these commits against https://github.com/whatwg/fs and create a PR there?

a-sully:
This issue has been ported to https://github.com/whatwg/fs/pull/10

PR #321, opened 2021-08-09
Add instructions for AccessHandle prototype

fivedots:
Adds the "Trying It Out" section, with pointers to the feature flag and the Emscripten file system.


rstz (approved):
LGTM

PR #354, opened 2022-01-20
Make clearer what is a `FileSystemFileHandle` and what a `FileSystem(Sync)AccessHandle`

tomayac:
Naming is mentioned as an [open question](https://github.com/WICG/file-system-access/blob/main/AccessHandle.md#naming), but I feel like this renaming helps people follow that it's a handle on a handle.


tomayac:
@fivedots, could you review this PR so @mkruisselbrink can merge? Thanks!

fivedots (approved):
LGTM, thanks!

PR #357, opened 2022-01-24
Fix drag and drop example

### LINKED ISSUE #356: Bug in example in specification
https://wicg.github.io/file-system-access/#drag-and-drop

CleanShot 2022-01-21 at 15 35 56@2x

As I mentioned about a web.dev example - https://twitter.com/tmcw/status/1481649674708676608 - this example includes a bug. If there are multiple files, the `await` in this loop will cause the next iteration of the loop to start on the next tick, and by then the dataTransfer will be disconnected. Working around this requires copying the dataTransfer.items array or saving the resolution of those Promise values to a Promise.all.

tomayac:
Fixes #356

***
<a href="https://pr-preview.s3.amazonaws.com/tomayac/native-file-system/pull/357.html" title="Last updated on Jan 24, 2022, 10:25 AM UTC (f101295)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/357/840b322...tomayac:f101295.html" title="Last updated on Jan 24, 2022, 10:25 AM UTC (f101295)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1801,8 +1801,8 @@ elem.addEventListener('drop', async (e) => {
   e.preventDefault();
 
   const fileHandlesPromises = [...e.dataTransfer.items]
-    .filter((item) => item.kind === 'file')
-    .map((item) => item.getAsFileSystemHandle());
+    .filter(item => item.kind === 'file')
+    .map(item => item.getAsFileSystemHandle());
 
   for await (const handle of fileHandlesPromises) {
     if (handle.kind === 'directory') {
--- a/index.bs
+++ b/index.bs
@@ -1796,20 +1796,19 @@ elem.addEventListener('dragover', (e) => {
   // Prevent navigation.
   e.preventDefault();
 });
+
 elem.addEventListener('drop', async (e) => {
-  // Prevent navigation.
   e.preventDefault();
 
-  // Process all of the items.
-  for (const item of e.dataTransfer.items) {
-    // kind will be 'file' for file/directory entries.
-    if (item.kind === 'file') {
-      const entry = await item.getAsFileSystemHandle();
-      if (entry.kind === 'file') {
-        handleFileEntry(entry);
-      } else if (entry.kind === 'directory') {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1796,20 +1796,19 @@ elem.addEventListener('dragover', (e) => {
...
-    if (item.kind === 'file') {
-      const entry = await item.getAsFileSystemHandle();
-      if (entry.kind === 'file') {
-        handleFileEntry(entry);
-      } else if (entry.kind === 'directory') {
-        handleDirectoryEntry(entry);
-      }
+  const fileHandlesPromises = [...e.dataTransfer.items]
+    .filter((item) => item.kind === 'file')
+    .map((item) => item.getAsFileSystemHandle());
```
kenchris, line 1805: you don't need () around item when there is just one argument. ```.map(item => item.getAsFileSystemHandle());``` works fine

tomayac: It's a style guide thing. Google highly [recommends](https://google.github.io/styleguide/jsguide.html#:~:text=tip%3A%20always%20using%20parentheses%20even%20for%20single-parameter%20arrow%20functions%20can%20avoid%20situations%20where%20adding%20parameters%2C%20but%20forgetting%20to%20add%20parentheses%2C%20may%20result%20in%20parseable%20code%20which%20no%20longer%20works%20as%20intended.) adding braces.

kenchris: I see. I usually leave them out as they just add more noise and make the code less readable in my opinion :-)

tomayac: This is the spec, the Google Style Guide doesn't apply here. Feel free to remove, I don't have hard feelings. It's a prettier rule that I have in muscle memory :-)

tomayac, line 1805:
```suggestion
    .filter(item => item.kind === 'file')
    .map(item => item.getAsFileSystemHandle());
```

mkruisselbrink (approved):

PR #366, opened 2022-03-23
Mark options parameter as required

fivedots:
Make the options parameter for createSyncAccessHandle and createAccessHandle required, to make it clear that a default behavior has not been decided yet.

Add a note about this on "Exposing AccessHandles on all filesystems", mentioning that the parameter should be made optional eventually.

### DIFF
--- a/AccessHandle.md
+++ b/AccessHandle.md
@@ -190,9 +190,9 @@ This proposal only currently considers additions to OPFS, but it would probably
 be worthwhile to expand the new functionality to arbitrary file handles. While
 the exact behavior of *AccessHandles* outside of OPFS would need to be defined
 in detail, it's almost certain that the one described in this proposal should
-not be the default. To avoid setting it as such, we propose adding a *mode*
-string parameter to *createAccessHandle()* and *createSyncAccessHandle()*.
-Some possible values *mode* could take are:
+not be the default. To avoid setting it as such, we propose adding a required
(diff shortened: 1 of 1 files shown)


AccessHandle.md
```diff
@@ -190,9 +190,9 @@ This proposal only currently considers additions to OPFS, but it would probably
 be worthwhile to expand the new functionality to arbitrary file handles. While
 the exact behavior of *AccessHandles* outside of OPFS would need to be defined
 in detail, it's almost certain that the one described in this proposal should
-not be the default. To avoid setting it as such, we propose adding an optional
-*mode* string parameter to *createAccessHandle()* and
-*createSyncAccessHandle()*. Some possible values *mode* could take are:
+not be the default. To avoid setting it as such, we propose adding a *mode*
+string parameter to *createAccessHandle()* and *createSyncAccessHandle()*.
+Some possible values *mode* could take are:
```
tomayac, line 195:
```suggestion
not be the default. To avoid setting it as such, we propose adding a required
*mode* string parameter to *createAccessHandle()* and
*createSyncAccessHandle()*. Some possible values *mode* could take are:
```

tomayac (approved):
LGTM, with one proposal to explicitly say the parameter is required (for now?).

a-sully (approved):
LGTM. I'll mention https://github.com/whatwg/fs/issues/19 here since this is closely related

PR #367, opened 2022-04-05
Remove options parameter from constructor

fivedots:
In order to prevent compatibility issues with WebKit, the options
parameter should be removed from `create[Sync]AccessHandle()`.

For more context see https://github.com/whatwg/fs/issues/19.


a-sully (approved):
LGTM

PR #384, opened 2022-08-24
Add note that suggestedName extension restrictions should match FilePickerOptions.types

a-sully:
Having these restrictions be different seems to go against user/developer expectations.

Extensions suggested via `suggestedName` should also not be longer than 16 code points, for example.


a-sully:
@mkruisselbrink friendly ping (trying to clean up open issues/PRs)

PR #390, opened 2022-11-11
Editorial: Fix references to permissions spec

a-sully:
Stolen from https://github.com/WICG/file-system-access/commit/dc8b56dc01ee3ad8bbb63ecd804b5be79c13262a. The changes to the permissions spec mentioned there have since been merged.

Also cleans up some trailing whitespace that we might as well clean up. Thanks, new editor settiing.

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/file-system-access/pull/390.html" title="Last updated on Nov 15, 2022, 12:06 AM UTC (d15bf6d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/390/70f0a00...a-sully:d15bf6d.html" title="Last updated on Nov 15, 2022, 12:06 AM UTC (d15bf6d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -203,7 +203,7 @@ permission-related algorithms and types are defined as follows:
   Given a {{FileSystemPermissionDescriptor}} |desc| and a {{PermissionStatus}} |status|,
   run these steps:
 
-  1. Run the [=boolean permission query algorithm=] on |desc| and |status|.
+  1. Run the <a spec=permissions>default permission query algorithm</a> on |desc| and |status|.
   1. If |status|.{{PermissionStatus/state}} is not {{PermissionState/"prompt"}}, abort.
   1. Let |settings| be |desc|.{{FileSystemPermissionDescriptor/handle}}'s [=relevant settings object=].
   1. Let |global| be |settings|'s [=environment settings object/global object=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1741,7 +1741,7 @@ these steps:
   1. Perform the <a spec=html>activation notification</a> steps in |global|'s [=Window/browsing context=].
 
   1. [=Request permission to use=] |desc|.
-  1. Run the [=boolean permission query algorithm=] on |desc| and |status|.
+  1. Run the <a spec=permissions>default permission query algorithm</a> on |desc| and |status|.
```
a-sully, line 1744: @mkruisselbrink this instance was not updated in https://github.com/WICG/file-system-access/commit/dc8b56dc01ee3ad8bbb63ecd804b5be79c13262a. I assume that was not intentional?

index.bs
```diff
@@ -203,7 +203,7 @@ permission-related algorithms and types are defined as follows:
   Given a {{FileSystemPermissionDescriptor}} |desc| and a {{PermissionStatus}} |status|,
   run these steps:
 
-  1. Run the [=boolean permission query algorithm=] on |desc| and |status|.
+  1. Run the <a spec=permissions>default permission query algorithm</a> on |desc| and |status|.
```
mkruisselbrink, line 206: Do these need the explicit spec= to disambiguate, or does it link fine without that as well (in which case it could also go back to the `[= ... =]` shorthand)?

index.bs
mkruisselbrink: I can't imagine it was, no.

mkruisselbrink (approved):
lgtm, with a question/nit

index.bs
a-sully: Ah good catch. The shorthand works

Issue #358, opened 2022-01-28
Remote file access (external cloud storage)


jimmywarting:
I think it would be really cool if there where any possibility to hook into cloud storages somehow...

I'm thinking sftp, WebDAV, smb like access... We could define a hole new protocol from the ground up, maybe even based on CORS policy. 

Imagine being able to hook into `Dropbox` or `Google Drive` and not having to worry about how some complicated SDK of how they handle file/directory handles, permissions, and walking the file tree structure etc. 

it would be super cool if we could define a web based protocol to communicate with cloud storages (possible with http2) with the same set of permission request model with reading/writing/seeking in a "standard" way of doing things.

I'm pretty sure it could be possible to get `File` class that isn't represented by some memory data or a filesystem path to some physical place on your own harddrive.
There is already something analogous in the http2 implementation in the form of the `respondWithFile()` and [respondWithFD()](https://nodejs.org/dist/latest-v17.x/docs/api/http2.html#http2streamrespondwithfdfd-headers-options) APIs in the http2 side.
so maybe it could be possible to get a `File` object that yet haven't been downloaded?
getting a file with `remoteHandle.getFile()` would only get metadata of the file such as filename, size and type somehow without actually downloading the content of the file. only when you create some reader/writer would you actually perform some upload/download.

```js
const directoryHandle = await navigator.storage.getRemoteDirectory('https://example.com/fsa/users/003/')
const directoryHandle = await navigator.storage.getRemoteDirectory('/fsa/users/003/')
const directoryHandle = await navigator.storage.getRemoteDirectory('sftp://address:port/path')
```

bradisbell:
Something similar can be done today with Service Workers, but unfortunately implementers have chosen to assume Service Workers are only useful for caching.  Therefore, when using various privacy modes or caching disabled, the Service Workers are disabled.

jimmywarting:
@bradisbell you sparked my interest with _"can be done today with Service Workers"_

how would you be able to get (say a) FileSystemDirectory handle from using a service worker and something coming from the service worker/network?

josephrocca:
@photopea has expressed a desire for something along these lines: https://github.com/photopea/photopea/issues/2118#issuecomment-636505499

photopea:
Since the File System Access API is asynchronous, any service can create an interface (a Javascript library) to provide their files under a File System Access API. 

[...]

PR #411, opened 2023-04-25
Add cloud file handling proposal

alex292:
This proposal adds a a new `getCloudHandles()` method to `FileSystemHandle`, which allows to retrieve cloud handles for a file/directory. A cloud handle consists of a vendor identifiert (e.g. "drive.google.com") and a file identifier. With these, the web app can talk to the cloud storage provider through its APIs directly to retrieve/modify the file. This is useful for web apps to figure out if a file is already backed by cloud storage and allows for easier transfer across machines as just the identifier instead of the entire file contents need to be transfered.


josephrocca:
Prior discussions that are related:

* https://github.com/WICG/file-system-access/issues/358
* https://bugs.chromium.org/p/chromium/issues/detail?id=1337347 - "Remote File Access with standard API to Cloud Storage providers"

jaime-rivas:
Hi @alex292,

I'm wondering if you'd be open to extend the proposal to include the non-goals as goals, specifically provide a standardized way for CSP’s web APIs to interact with remote files.

There is another issue opened here #358 and in [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1337347) where we briefly talk about those goals. It seems like a great opportunity to finally integrate cloud storage into browsers to give standardized APIs to webapps to interact with remote files and remove dependency from the underlying operating system. Thanks in advance!

alex292:
That is currently not part of our considerations as that would not solve any of the listed use cases, so the answer is sadly "no".
I would also argue that unifying the CSP APIs should not be done via a web API in a browser. The same way you would need to include some form of SDK from each specific CSP vendor to interact with their APIs, there might be one SDK that solves this use case by combining these into a common interface. The browser would not be involved in any of this and therefore it should also not be a web API IMHO.

josephrocca:
I guess the ideal user experience here might be that the OS/user-agent stores details of the user's cloud storage accounts, and then simply presents those folders as options in the `showOpenFilePicker` dialogue.

That way it's completely up to the user and the user agent which filesystem is being interacted with (regardless of whether it's remote or local, or remote-but-locally-cached) - and the developer doesn't need to care about it at all. They just treat their file/directory handles like they normally would.

proposals/CloudFileHandle.md
```diff
@@ -30,10 +30,10 @@ if(cloudHandles.length === 0) {
   // File is not synced by any CSP
 }
 for(const cloudHandle of cloudHandles) {
-  if(cloudHandle.provider === 'drive.google.com') {
```
jimmywarting: it normally never follows the `provider_name` pattern. 
everywhere else it's named camelCase

so maybe should be `providerName`

alex292: good catch, thanks!

jimmywarting:
Hmm, honestly this is not what i had in mind when i created #358
i have multiple cloud providers, like google drive and dropbox, but i don't have any cloud provider syncing the data with with any of my devices.

i only keep them in the cloud ☁️
(Only exception is for my android phone that uploads my photos to google photo to have backup)

So when i asked for #358, then this:
```js
const [fileHandle] = await window.showOpenFilePicker(pickerOpts);
const cloudIdentifiers = await fileHandle.getCloudIdentifiers();
```
...was not what i had in mind.

I kind of just wished that there where some magical way a website that you have visited before could just magically register itself as a cloud provider (Maybe with a manifest.json + service worker?)

And the next time you called `showOpenFilePicker()` then you would have the option to choose to pick files directly from google drive. 

my idea was for developer to not being able to know weather or not it's a local file handle, or a remote-but-locally-cached or a fully-remote file handle

the idea was to never have to install any desktop/phone application and just straight up being able to talk to services that you are logged in to in your browser.

alex292:
This proposal was not meant to solve https://github.com/WICG/file-system-access/issues/358, I have actually never seen that proposal before.

With this new web API, we are trying to solve the use cases discussed in the "use cases" section. I have been working with asully@chromium.org on this so far.

jimmywarting:
I'm still confused as to why we even need this `getCloudHandles()` method... 
I don't think i would ever use it... 
I rather wished i got a normal looking `FileSystemHandle` and that i had no idea if i where writing things directly to the cloud as if i had been granted access to `ftp://usr:psw@drive.google.com/pic/me.png` or just a local file in the disk.

My intention with #358 was that you should never have to do any code specific things more dedicated towards google drive, dropbox, or OneDrive. 

The hole point of #358 was to avoid having to learn how to use a Google Drive specific API and doing things like
```js
  if (cloudIdentifier.providerName === 'drive.google.com') {
    // retrieve/modify the file from Google Drive API using cloudIdentifier.id
  }
 ```
 then you must register a developer account at Google, get some kind of API key / token or something like that.
learn how how google's api / SDK works if you wish to do anything remotly useful with this file handle.

And the lightly hood that you get a cloud Identifier that your specific web app could understand and talk directly to is more unlikely. there exist hundred of different cloud provider. so every site would have to ad here to learn how to use each and every possible cloud api, where as #358 was trying to normalize the way of talking directly to a cloud provider without having to deal with lots of different code paths that could be taken. 

Now if the usecase is to be able to easily share something that's already uploaded to a cloud provider wouldn't it then be easier to just use the [Web Share API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API) or something like that instead? allow the Web Share API to accept a FileSystemHandle? maybe it will instead share a link rather than uploading file.

---

learning how to use git, ftp, sftp WebDav, or smb is way easier to learn and that applies to multiple servers and it is way easier to learn then it's to having to dig deep into how any cloud specific api such as the Google Drive API works.

it's google drive, dropbox and others who should have to learn how to hook in their own cloud system into the OS as a network drive, it should not be the other way around.

**Comment by jimmywarting** [comment]
> Imagine the file/directory you want to share being 5GB large, you likely don't want to upload that from your device, but instead quickly pull it to the remote machine from Google/Microsoft servers directly.

I see your point that taking something from google photos and uploading it directly to eg facebook would be a valid point in that you would not have to download it from google and then re-upload it to facebook.

But honestly i think that is something that could maybe technically be solved with WebShare API where sharing a remote `FileSystemHandle` could just be turned up as a download link or something.

A hole other solution could be to create something serializable like `FileSystemHandle.getUniqueId()` whatwg/fs#46 so they can be share to facebook so it could reconstruct a FileSystemHandle in there backend (server) and then fb could pull the data directly from google drive by calling something like

it would be like handing dropbox a ftp link saying "here a uniq FileSystemHandle identifier that happens to look something like `ftp://usr:uniqtoken@drive.google.com/my-summercamp-2022` go and call `FileSystemFileHandle.from('ftp://usr:uniqtoken@drive.google.com/my-summercamp-2022')` to reconstruct it and pull it directly from google instead.

File: proposals/CloudIdentifier.md [hunk 2]
```diff
@@ -0,0 +1,197 @@
...
+
+### Non-Goals
+
+This proposal does not plan to provide a standardized way for CSP’s web APIs to 
+* give access permissions to these files/directories.
+* interact (fetch/modify/etc) with these files/directories.
+* provide additional meta-data (e.g. sync status)
+
+Instead, it should only generate an opaque identifier for a file/directory so that web apps can then interact with these individual CSPs’ web APIs on the same file/directory. Obtaining the required access permissions on that file is delegated to the web app.
+Web apps might only support one specific CSP. When supporting multiple CSPs, each CSP will require their own implementation, although semantically they would likely perform similar steps.
```
a-sully, line 94: By this I assume you mean that a web app might only know how to talk to, say, Google Drive, but not OneDrive?

The number of CSPs that a web app knows how to talk with seems like an implementation detail of that app, and not all that relevant from the perspective of the `getCloudIdentifiers()` web API? The web API only cares what CSPs from the local machine have registered themselves with the user agent - not which of of those CSPs the calling web app knows how to talk with

Do you think it could be worth making more explicit that this proposal is really just to allow the user agent to act as a bridge between the local CSP client and the CSP server, and that it _does not_ handle the actual reading/writing of files with the CSP? (as summarized nicely in this comment https://github.com/WICG/file-system-access/issues/358#issuecomment-1522280549)

proposals/CloudIdentifier.md
```diff
@@ -0,0 +1,197 @@
...
+
+![image](./images/cloud-identifier/design.png "Design")
+
+1. Locally installed CSP client registers itself as a provider for certain directories with the browser
+2. Web app receives `FileSystemHandle` for a file/directory (via File Handling opening flow, File System Access file/directory picker or drag & drop) and requests its cloud identifier(s) via `getCloudIdentifiers()`.
+3. Browser checks whether this path has been registered as being synced by a provider
+   * If no, resolve promise with empty list
+4. For each registered provider for that path, the browser will send a request to the registered provider’s executable with the file’s/directory’s path
+5. Provider requests a token for that path from the CSP
+   * The token can also be a stable and already cached identifier, in which case step 5 and 6 are skipped
```
a-sully, line 106: This feels like an implementation detail of the CSP that might be too specific to belong here? Thoughts?

proposals/CloudIdentifier.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Fingerprinting
+
+The browser has no control whether the CSP will provide one-time tokens or stable identifiers as part of their cloud identifier. If the CSP provides stable identifiers, the web application could use these as a [fingerprinting](https://www.w3.org/TR/fingerprinting-guidance/#dfn-active-fingerprinting) mechanism for the files/directories it has access to.
+
+In theory, if a web application already has access to a `FileSystemHandle` the web application could already use other mechanisms to generate fingerprinting identifiers, but these are less stable:
+* Hashing the file’s/directory’s contents -> identifier will change if file/directory content changes
+* `FileSystemHandle::getUniqueId()` [[explainer](https://github.com/whatwg/fs/pull/46)] -> clearing browsing data will reset unique IDs
+
+The web application would need repeated access to the same `FileSystemHandle` to perform this fingerprinting though. That means the user must either re-grant access to the same file or the web app stores the file handle in an IndexDB, which can be cleared by the user though by clearing their browsing data.
```
a-sully, line 181: nit: "IndexedDB"

proposals/CloudIdentifier.md
```diff
@@ -0,0 +1,197 @@
...
+
+It would also be up to the CSP whether they actually provide permanent tokens or temporary tokens.
+
+### Modification via read-only permission
+
+If a web application only has `read` level access to a `FileSystemHandle`, but has `write` level access to that file via CSP web APIs, it could still modify the cloud-stored file, which is then synced to the device and thereby modify the file.
```
a-sully, line 187: nit: `"read"` and `"readwrite"` (and feel free to link to https://wicg.github.io/file-system-access/#enumdef-filesystempermissionmode)

proposals/CloudIdentifier.md
```diff
@@ -0,0 +1,197 @@
...
+
+The web application would need repeated access to the same `FileSystemHandle` to perform this fingerprinting though. That means the user must either re-grant access to the same file or the web app stores the file handle in an IndexDB, which can be cleared by the user though by clearing their browsing data.
+
+It would also be up to the CSP whether they actually provide permanent tokens or temporary tokens.
+
+### Modification via read-only permission
+
+If a web application only has `read` level access to a `FileSystemHandle`, but has `write` level access to that file via CSP web APIs, it could still modify the cloud-stored file, which is then synced to the device and thereby modify the file. 
+
+In this case, the user has clearly granted write access to the CSP-backed file and by having a sync client, the user also allows the local files to be modified, so the change would actually not be surprising.
```
a-sully, line 189: I would emphasize that this is possible today with existing APIs. From the perspective of the web platform APIs, the website is just writing to a standard file on disk and it just so happens that the changes are being synced by the CSP

proposals/CloudIdentifier.md
```diff
@@ -0,0 +1,197 @@
...
+
+1. Locally installed CSP client registers itself as a provider for certain directories with the browser
+2. Web app receives `FileSystemHandle` for a file/directory (via File Handling opening flow, File System Access file/directory picker or drag & drop) and requests its cloud identifier(s) via `getCloudIdentifiers()`.
+3. Browser checks whether this path has been registered as being synced by a provider
+   * If no, resolve promise with empty list
+4. For each registered provider for that path, the browser will send a request to the registered provider’s executable with the file’s/directory’s path
+5. Provider requests a token for that path from the CSP
+   * The token can also be a stable and already cached identifier, in which case step 5 and 6 are skipped
+6.  CSP can choose to generate a one-time token or stable identifier to later be used by the CSP’s web APIs. Either way, the properties of the token are up to the CSP and completely opaque to the browser.
+7. Provider responds to browser with the token
```
a-sully, line 108: Meta comment: this explainer might be easier to follow if the terminology was more consistent. For example, "CSP" and "provider" are mostly used interchangeably, though for someone not intimately familiar with the problem space it might not be clear that these refer to the same thing (not to mention that a "CSP" itself has multiple parts - the local sync client and the server)

a-sully (changes requested):

proposals/CloudIdentifier.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### Drag & Drop into Mail
+
+Sharing files via mail used to be done using simple file attachments. Especially for large files, it is preferable to instead upload them to a CSP and then share an access link to that file to the recipient. Google Mail’s web application, for example, already does this by prompting the user to upload a file to Google Drive when attaching a large file.
+
+With the proposed changes, the web application could already detect that the file is synced by cloud storage and generate a share link for that file’s cloud identifier without requiring the user to manually go through these steps or prompt the user to upload the file to cloud storage again.
+
+### Non-Goals
+
+This proposal does not plan to provide a standardized way for CSP’s web APIs to
```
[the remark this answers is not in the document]
alex292: rewritten the sentence

proposals/CloudIdentifier.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### Remote file handling
+
+Web applications offering app streaming or VDI might want to make a locally synced file available to the remote virtual machine where the app/desktop is being streamed from. For example, the user runs their image editor on a remote machine, launches the file picker to open image.png from the local client device, makes some changes and then saves the file with Ctrl+S.
+
+##### Before
+
+![image](./images/cloud-identifier/remote-file-handling-before.png "Before")
+
+1. VDI web app requests and receives the file via `FileSystemFileHandle::getFile()`
```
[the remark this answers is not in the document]
alex292: Done

proposals/CloudIdentifier.md
```diff
@@ -0,0 +1,197 @@
...
+5. VDI web app writes updated file contents via `FileSystemFileHandle::write()`, which is picked up by the CSP client app
+6. CSP client app synchronizes the file by uploading it to the CSP server
+
+In this scenario all transfers (arrows in diagram) transfer the entire file and the file has to be downloaded and uploaded from the local device’s network connection.
+
+##### After
+
+![image](./images/cloud-identifier/remote-file-handling-after.png "After")
+
+1. VDI web app requests and receives the file’s cloud identifier via FileSystemFileHandle::getCloudIdentifiers()
```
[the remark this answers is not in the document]
alex292: Done

proposals/CloudIdentifier.md
alex292: rewritten this part a bit to make it more clear. It's a bit weird calling `getCloudIdentifiers()` and the CSPs' APIs both "web API"s. One is a web API in the browser, while the other one is an API towards the CSP back-end. Any suggestions for better naming here?

proposals/CloudIdentifier.md
alex292: removed

proposals/CloudIdentifier.md
alex292: Done, should be more clear in regards to "local sync client" or "CSP backend" now

proposals/CloudIdentifier.md
alex292: Done

proposals/CloudIdentifier.md
alex292: Done

proposals/CloudIdentifier.md
alex292: replaced `readwrite` with "edit/write".

That is pretty much out of scope of this web API. It could very well be that the web app only has read permission on the CSP backend level. Not all use cases involve actually writing to the CSP backend.
Line 95 describes this delegation of getting the correct access levels.

cameyo:
This would be great to integrate into our product. In our virtual app delivery platform, we support filetype association. So for example when the user double-clicks a .psd file, we currently read it locally through the JS handle, send it over to a cloud-based execution host and open it there. Then, we sync all changes back and forth to the user's device.

With this functionality, we'd gain a lot of speed and network usage, as the cloud execution host server would be able to open it directly from the Drive API (initial speed / network saving), and then the same time saving would apply for change syncing. This gain would be particularly interesting with >1MB files, which is often the case.

As a good "side effect", this will also resolve for us sharing and exclusivity issues -- i.e. what happens today when the user works within a cloud session on a local file, and then that file is moved, deleted or changed locally.

We already integrate with the Drive API, so having to go through JS file handles is a bit awkward currently. Definitely a big +1 !

PR #423, opened 2023-06-23
Default FilePickerAcceptType.description to the empty string

### LINKED ISSUE #422: Default FilePickerAcceptType.description to the empty string
Not this PR, but it seems like we can make this "" (empty string) by default in the IDL, then we don't need this in prose?

_Originally posted by @inexorabletash in https://github.com/WICG/file-system-access/pull/419#discussion_r1238942688_

a-sully:
This has no behavioral changes, since _not specifying_ the description field currently behaves the same as if the description was specified as the empty string

Fixes #422

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/file-system-access/pull/423.html" title="Last updated on Jun 23, 2023, 4:08 PM UTC (7d53d65)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/file-system-access/423/739b0c4...a-sully:7d53d65.html" title="Last updated on Jun 23, 2023, 4:08 PM UTC (7d53d65)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -299,7 +299,7 @@ enum WellKnownDirectory {
 typedef (WellKnownDirectory or FileSystemHandle) StartInDirectory;
 
 dictionary FilePickerAcceptType {
-    USVString description;
+    USVString description = "";
     record<USVString, (USVString or sequence<USVString>)> accept;
 };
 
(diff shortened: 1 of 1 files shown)


a-sully:
FYI the implementation is updated here: https://crrev.com/c/4638360

Although the IDL is changing, can you confirm that we don't need to go through any sort of launch process because it's not actually a web-exposed change?

inexorabletash (approved):

inexorabletash:
Re: launch process - that's a question for the Chromium change. :) But yes, if it's not web-visible and the reviewers agree it's sufficiently tested etc, no process should be needed.

a-sully:
> Re: launch process - that's a question for the Chromium change. :) But yes, if it's not web-visible and the reviewers agree it's sufficiently tested etc, no process should be needed.

SGTM

Issue #131, opened 2019-11-20
In close(), what counts as "file on disk changed"?


foolip:
[`close()`](https://wicg.github.io/native-file-system/#api-filesystemwriter-close) is a TODO for normative text, and one thing I'd expect to need quite a lot of detail is what counts as "file on disk changed". It could be one of:
- file name or content is currently not what it was when it was opened
- file name or content was different some time between now and when it was opened
- file modification time is not the same as when it was opened, although file contents might be the same

mkruisselbrink:
Close does have normative text now (threaded through writable streams), and we actually didn't implemented (nor specified) the described behavior. Not sure if we should bring something like this back though. It does seem like a potentially nice backstop against data loss, but it might also result in unexpected (and hard to debug) errors for developers...

PR #427, opened 2023-08-14
Update security questionnaire to current version

nathanmemmott:
Updates each question to use the current version's wording. Answers questions that were added, and removes ones that were removed.

### DIFF
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -1,4 +1,4 @@
-https://www.w3.org/TR/2019/NOTE-security-privacy-questionnaire-20190523/
+https://www.w3.org/TR/2021/NOTE-security-privacy-questionnaire-20211216/
 
 ### 2.1. What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
 
@@ -6,19 +6,19 @@ This feature exposes files and directories the user explicitly selects to share 
 
 However this feature does make it possible for browsers to extend the time web sites have access to this information. I.e. permission grants and handles received through this API could be persisted giving web sites access to the same files/directories on disk when a user later returns to the website (or when a service worker for the origin is processing an event). At least for the Chrome implementation, we're only planning on having these grants persist for installed PWAs. On the drive-by web, access will only be as long as the web site is open, requiring re-prompting in subsequent visits.
 
-### 2.2. Is this specification exposing the minimum amount of information necessary to power the feature?
+### 2.2. Do features in your specification expose the minimum amount of information necessary to enable their intended uses?
 
 Yes, we're only exposing files and directories explicitly selected by the user. Of course, web sites could ask for access to a directory when all it needs is access to some files, but the same is already true today. At least as far as exposing information to web sites is concerned, this API doesn't expose any more information than existing APIs do today.
 
-### 2.3. How does this specification deal with personal information or personally-identifiable information or information derived thereof?
+### 2.3. How do the features in your specification deal with personal information, personally-identifiable information (PII), or information derived from them?
 
 No data is exposed without the user explicitly choosing what files or directories to expose to the web site.
 
-### 2.4. How does this specification deal with sensitive information?
+### 2.4. How do the features in your specification deal with sensitive information?
 
 No data is exposed without the user explicitly choosing what files or directories to expose to the web site, so only sensitive data that the user explicitly decides to share via this API will be shared. Furthermore, this API is only exposed in secure contexts, and third-party iframes (i.e. iframes that are cross origin from the top-level frame) won't be able to show pickers or permission prompts and can only access data they were already granted access to from a top-level same origin frame.
 
-### 2.5. Does this specification introduce new state for an origin that persists across browsing sessions?
+### 2.5. Do the features in your specification introduce new state for an origin that persists across browsing sessions?
 
 Yes, this specification lets websites store handles they've gotten access to (via a file or directory picker) in IndexedDB. User agents could also persist the permission grants that go with these handles, but at least in the Chrome implementation, these permission grants will only be persistent for installed PWAs. The drive-by web will only have enough state to allow it to re-prompt for access, but the access itself won't be persistent.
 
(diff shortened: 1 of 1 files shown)


security-privacy-questionnaire.md
```diff
@@ -1,24 +1,24 @@
-https://www.w3.org/TR/2019/NOTE-security-privacy-questionnaire-20190523/
+https://www.w3.org/TR/2021/NOTE-security-privacy-questionnaire-20211216/
 
 ### 2.1. What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
```
a-sully, line 3: This question now has sub-questions, requiring explicit consideration of four distinct "parties". Let's add a TODO to address this in a follow-up?

https://www.w3.org/TR/2021/NOTE-security-privacy-questionnaire-20211216/#purpose

security-privacy-questionnaire.md
```diff
@@ -1,24 +1,24 @@
...
 
 No data is exposed without the user explicitly choosing what files or directories to expose to the web site, so only sensitive data that the user explicitly decides to share via this API will be shared. Furthermore, this API is only exposed in secure contexts, and third-party iframes (i.e. iframes that are cross origin from the top-level frame) won't be able to show pickers or permission prompts and can only access data they were already granted access to from a top-level same origin frame.
 
-### 2.5. Does this specification introduce new state for an origin that persists across browsing sessions?
+### 2.5. Do the features in your specification introduce new state for an origin that persists across browsing sessions?
 
 Yes, this specification lets websites store handles they've gotten access to (via a file or directory picker) in IndexedDB. User agents could also persist the permission grants that go with these handles, but at least in the Chrome implementation, these permission grants will only be persistent for installed PWAs. The drive-by web will only have enough state to allow it to re-prompt for access, but the access itself won't be persistent.
```
a-sully, line 23: This text is outdated. Let's add another TODO to update this once we publish our proposal for https://github.com/WICG/file-system-access/issues/297?

security-privacy-questionnaire.md
```diff
@@ -30,50 +30,55 @@ Additionally, user agents could also choose to persist the last directory a file
...
 
-### 2.11. Does this specification allow an origin some measure of control over a user agent’s native UI?
+### 2.11. Do features in this specification allow an origin some measure of control over a user agent’s native UI?
 
 The origin can pop up native file or directory pickers, and have some control over what appears inside that native UI (e.g. accepted file types, starting directory and suggested file names), but that control is very limited. The spec does put limitations on what is allowed as an accepted file type and suggested file name to limit the security impact of allowing websites to control the native UI. User agents are expected to employ similar mechanisms to sanitize the sugessted file names as are used to sanitize for example suggested file names in `<a download="foo.ln">` today.
 
-### 2.12. What temporary identifiers might this this specification create or expose to the web?
+### 2.12. What temporary identifiers do the features in this specification create or expose to the web?
```
a-sully, line 56: @alex292 would you like to create a follow-up PR adding a note about https://github.com/WICG/file-system-access/blob/main/proposals/CloudIdentifier.md ?

security-privacy-questionnaire.md
```diff
@@ -1,24 +1,24 @@
...
 
 No data is exposed without the user explicitly choosing what files or directories to expose to the web site.
 
-### 2.4. How does this specification deal with sensitive information?
+### 2.4. How do the features in your specification deal with sensitive information?
 
 No data is exposed without the user explicitly choosing what files or directories to expose to the web site, so only sensitive data that the user explicitly decides to share via this API will be shared. Furthermore, this API is only exposed in secure contexts, and third-party iframes (i.e. iframes that are cross origin from the top-level frame) won't be able to show pickers or permission prompts and can only access data they were already granted access to from a top-level same origin frame.
 
-### 2.5. Does this specification introduce new state for an origin that persists across browsing sessions?
+### 2.5. Do the features in your specification introduce new state for an origin that persists across browsing sessions?
```
a-sully, line 21: This question gives IndexedDB as an example... which suggests the BucketFS should be included here. Or it might be worth creating a separate questionnaire on the whatwg spec?... Anyways, let's just add a TODO for now?

https://www.w3.org/TR/2021/NOTE-security-privacy-questionnaire-20211216/#persistent-origin-specific-state

security-privacy-questionnaire.md
```diff
@@ -30,50 +30,55 @@ Additionally, user agents could also choose to persist the last directory a file
...
+### 2.12. What temporary identifiers do the features in this specification create or expose to the web?
 
 The `getUniqueId()` method will create a temporary unique identifier for a given handle. This ID will become invalid if the user clears storage for the site.
 
 ### 2.13. How does this specification distinguish between behavior in first-party and third-party contexts?
 
 It is expected that user agents do not allow third-party contexts to prompt for any kind of access using this API. I.e. third-party contexts can potentially access files or directories that their origin was already granted access to in a first-party context (by sharing handles via IndexedDB or postMessage), but can't trigger any new file/directory pickers or permission requests.
 
-### 2.14. How does this specification work in the context of a user agent’s Private Browsing or "incognito" mode?
+### 2.14. How do the features in this specification work in the context of a browser’s Private Browsing or Incognito mode?
```
a-sully, line 64: If we decide to include discussions of the BucketFS in this questionnaire (in a follow-up, presumably), we should mention that it's expected to be implemented using an in-memory file system in this case

security-privacy-questionnaire.md
```diff
@@ -30,50 +30,55 @@ Additionally, user agents could also choose to persist the last directory a file
...
 Yes.
 
-### 2.16. Does this specification allow downgrading default security characteristics?
+### 2.16. Do features in your specification enable origins to downgrade default security protections?
 
 No.
 
+### 2.17. How does your feature handle non-"fully active" documents?
+
+This feature uses a locking system to prevent data races when modifying/reading a handle. Locks are acquired on the handle. These locks are held by the user agent, not the platform.
```
a-sully, line 78: > Locks are acquired on the handle.

Well, [not exactly](https://github.com/whatwg/fs/pull/131)... This doesn't add much IMHO so I would recommend leaving it out

> These locks are held by the user agent, not the platform.

[Not necessarily](https://crbug.com/1240780). This is up to the user agent's implementation and is subject to change. Should probably also leave this out

security-privacy-questionnaire.md
```diff
@@ -30,50 +30,55 @@ Additionally, user agents could also choose to persist the last directory a file
...
+
+When lock contention occurs between a "fully active" and non-"fully active" document, the latter is discarded and its locks released.
```
a-sully, line 80: Prior to this sentence, we should mention explicitly that a page may continue to hold a lock after being BFCached

a-sully (commented):
Mostly LGTM. Thanks for making these changes!

security-privacy-questionnaire.md
alex292: Happy to do so. Given that these identifiers are mostly not *temporary*, is 2.12 the right location for this or should this rather go into 2.5?
Probably also 2.6, right?

a-sully: > Happy to do so. Given that these identifiers are mostly not _temporary_, is 2.12 the right location for this or should this rather go into 2.5?

Good point. Let's do 2.5, since the characteristics of this identifier is up to the CSP's discretion

> Probably also 2.6, right?

Yup, good catch!

a-sully:
FYI merge conflicts are due to https://github.com/WICG/file-system-access/commit/dd9f1f9fddc6cf66c0250314012251cfd31b6650

PR #443, opened 2024-03-19
showDirectoryPicker always returns 1 handle

vapier:


tomayac (approved):
LGTM, approval

vapier:
that's kind of a lot of red tape.  i can see why some people would not bother contributing.  but i guess that's not specific to this API proposal.

i logged in with my account (that i luckily have) and smashed a lot of buttons.  guess we'll see what happens.

a-sully:
> that's kind of a lot of red tape

Ah yes, I just remembered this button, which applies here

> [a-sully](https://github.com/a-sully) marked as non substantive for IPR from ash-nazg.

